PR #19, opened 2016-06-28
Adding CMake build

mijamo911:


ionc/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  ion_writer.c
+  ion_writer_text.c
+)
+set_property(TARGET ionc PROPERTY C_STANDARD 99)
+target_include_directories(ionc PUBLIC
+  .
+  inc/
+
+)
+target_link_libraries(ionc decNumber m)
```
mijamo911: "m" is the math library, which isn't linked be default for historical reasons

ionc/ion_reader_text.c
```diff
@@ -651,8 +655,6 @@ enum version_marker_result _ion_reader_text_parse_version_marker(ION_STRING* ver
     return SUCCESS;
 }
 
-extern inline int add_digit(int i, char digit);
```
mijamo911: Switched this guy to inline static as originally intended

test/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_executable(tester
```
mijamo911: tester worked on the first try, hooray! All tests pass

We may want to model this as an actual CMake test, haven't done a whole lot of investigation but it appears we can pass args to an executable so we could just pass in the path to the ion-tests submodule to make it super easy

.gitignore
```diff
@@ -0,0 +1,6 @@
+CMakeCache.txt
+CMakeFiles/
+*/CMakeFiles/
+cmake_install.cmake
```
almann: Is this line redundant to the one below it?

.gitignore
```diff
@@ -0,0 +1,6 @@
+CMakeCache.txt
+CMakeFiles/
```
almann: Is this line redundant to the one below it?

almann:
Overall looks good--can we streamline the ignore file?

.gitignore
mijamo911: Not from my testing, both were required for the desired effect

PR #24, opened 2016-07-05
Adding basic framework for unit tests

mijamo911:


debug-tests
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+gdb --args build/debug/test/tester ion-tests
```
almann: What about platforms that don't look like gcc?

test/ion_assert.c
```diff
@@ -0,0 +1,5 @@
+#include "ion_assert.h"
+
+iERR assert_equals_int(int expected, int actual) {
+    return (expected == actual) ? IERR_OK : IERR_INVALID_ARG;
+}
```
almann: I'd just fold this into the header and make it `static inline` no sense for an extra file and translation unit for something so trivial.

almann:
When a test fails do I get some clue as to what broke.  I'm guessing I do as a result of `ION_CHECK`, but can you confirm?

Should we be using a C unit testing framework of some sort?  I am somewhat ambivalent since I don't like additional dependencies, but we should at least evaluate if any would be reasonable and make our lives easier.

debug-tests
mijamo911: No idea? I figure we'll cross the cross-platform bridge when we come to it, just getting the code to compile is a more likely first step than getting debugging working. Or did you mean just add "-gdb" to the file name so that it's clear what's going on?

I'll add the file extensions

test/ion_assert.c
mijamo911: Will do

debug-tests
almann: My point is there should be some kind of *TODO*, issue, or event a mention at a minimum that this is specific to a particular platform and we'll support other platforms in another way.

We don't need to support all the platforms, but we should be thinking about what we need to do for the ones we need to support.

mijamo911:
We already a have a (very rudimentary) unit test setup with counters for suites and tests, I'd prefer to augment that than take any additional dependencies. I've beefed up the assertions so that useful error messages will be output, and I've converted them to macros so that we can get accurate line number information.

test/ion_binary_test.c
```diff
@@ -0,0 +1,40 @@
...
+    run_unit_test(test_ion_binary_len_uint_64);
+    run_unit_test(test_ion_binary_len_int64);
+
+    iRETURN;
+}
+
+iERR test_ion_binary_len_uint_64() {
+    iENTER;
+
+    IONCHECK(ASSERT_EQUALS_INT(0, ion_binary_len_uint_64(0LL), "Wrong uint64 length"));
```
tgregg: Can the IONCHECK go inside the assertion?

tgregg:
Looks like a good start to me.

PR #25, opened 2016-07-06
Removing references to YJSDK

mijamo911:


almann:
Looks good--please add an issue for pluggable malloc/free if it doesn't already exist.

PR #27, opened 2016-07-08
Build improvements

hohle:
Creates the intermediate `build` directory before changing into it.

Conditionally includes `malloc.h` on Windows (`malloc(3)` is  part of `stdlib.h` in C99).

Tested on:

* OS X 10.11.5
* FreeBSD 10.3
* CentOS 7.2

### DIFF
--- a/tools/ionizer/options.c
+++ b/tools/ionizer/options.c
@@ -10,8 +10,8 @@
 #include <stdio.h>
 #ifdef WIN
 #include <io.h>
-#endif
 #include <malloc.h>
+#endif
 #include <string.h>
 #include <ctype.h>
 
--- a/tools/ionsymbols/options.c
+++ b/tools/ionsymbols/options.c
@@ -10,8 +10,8 @@
 #include <stdio.h>
 #ifdef WIN
 #include <io.h>
-#endif
 #include <malloc.h>
+#endif
 #include <string.h>
 #include <ctype.h>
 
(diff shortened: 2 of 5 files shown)


tools/ionizer/options.c
```diff
@@ -10,8 +10,8 @@
 #include <stdio.h>
 #ifdef WIN
 #include <io.h>
-#endif
 #include <malloc.h>
+#endif
```
almann: My understanding is that `malloc.h` is an archaic header for Linux systems, and is probably not on MSVC.  Perhaps we should eliminate this include all together or prefer `stdlib.h`

tools/ionsymbols/options.c
```diff
@@ -10,8 +10,8 @@
 #include <stdio.h>
 #ifdef WIN
 #include <io.h>
-#endif
 #include <malloc.h>
+#endif
```
almann: Same as above.

almann:
This looks good, thanks for doing this--I don't think `malloc.h` exists in MSVC, I think we can kill it.

tools/ionizer/options.c
hohle: I left it in based on the [MSDN `malloc` docs](https://msdn.microsoft.com/en-us/library/6ewkz86d.aspx), but I have no way to test it.

almann: Fair enough.

PR #30, opened 2016-07-14
Fixing a bunch of random issues to reduce warnings in build output

mijamo911:


ionc/inc/ion_int.h
```diff
@@ -97,7 +97,7 @@ typedef struct _ion_int {
     int       _signum;       // sign, +1 or -1, or 0
     SIZE    _len;          // number of digits in the _digits array (-1 if null)
     II_DIGIT *_digits;       // array of "digits" in some large base (2^31 currently)
-};
+} _ion_int;
```
mijamo911: Apparently the required syntax is typedef struct foo { ... } bar;

In cases where you don't care about bar, you can just set it equal to foo.

ionc/inc/ion_types.h
```diff
@@ -137,6 +137,7 @@ typedef struct _ion_collection          ION_COLLECTION;
 typedef struct _ion_stream        ION_STREAM;
 
 // decl's for user managed stream
+struct _ion_user_stream;
```
mijamo911: So what we're doing here is declaring the struct, referencing the declaration, and then defining it - all within a few lines. This is the idiomatic (well, maybe just correct) way to handle these kinds of circular definitions.

ionc/ion_binary.c
```diff
@@ -26,7 +26,7 @@ int ion_binary_len_uint_64(uint64_t value) {
 }
 
 // These field formats are always used in some context that clearly indicates the number of octets in the field.
-int ion_binary_len_int64(int64_t value) {
+int ion_binary_len_int_64(int64_t value) {
```
mijamo911: We exported a different function name in the header (int_64 vs int64), and the style chosen here matches the existing style in the other functions.

ionc/ion_reader_binary.c
```diff
@@ -368,8 +368,8 @@ iERR _ion_reader_binary_step_out(ION_READER *preader)
         ASSERT(preader->_eof ? (curr_pos == next_start) : (curr_pos <= next_start));  
         to_skip = next_start - curr_pos;
         while (to_skip > 0) {
-            if (to_skip > SIZE_MAX) {
-                one_skip = SIZE_MAX;
+            if (to_skip > MAX_SIZE) {
+                one_skip = MAX_SIZE;
```
mijamo911: to_skip is int64_t, and one_skip is SIZE which is int32_t. SIZE_MAX is a built-in macro in C which is of type int64_t, whereas MAX_SIZE is an Ion macro of type int32_t set to INT32_MAX. Since we were unintentionally downcasting and it's pretty obvious that our SIZE and MAX_SIZE macros are intended to commingle this was an easy fix

ionc/ion_reader_text.c
```diff
@@ -836,7 +836,7 @@ iERR _ion_reader_text_step_out(ION_READER *preader)
             sub_type = IST_LIST;
             break;
         default: {
-            char* error_message[ION_ERROR_MESSAGE_MAX_LENGTH];
+            char error_message[ION_ERROR_MESSAGE_MAX_LENGTH];
```
mijamo911: This is just a case of declaring char*[] (which is essentially char**) when we meant char[].

ionc/ion_stream.c
```diff
@@ -141,18 +141,21 @@
 #elif defined(ION_PLATFORM_WINDOWS)
   #define LSEEK _lseek
 #else
+  #include <unistd.h>
```
mijamo911: These blocks should probably be unified, but basically there's three separate checks to see if we're windows or not and then if we're Linux-ish we define some macro aliases for posix APIs but never actually include the proper headers.

ionc/ion_stream_impl.h
```diff
@@ -127,7 +127,6 @@ typedef uint32_t  ION_STREAM_FLAG;
 #define FLAG_CAN_WRITE          0x00200
 #define FLAG_RANDOM_ACCESS      0x00400
 #define FLAG_IS_TTY             0x00800
-#define FD_IS_TTY(fd)           _isatty(fd) /* TODO */
```
mijamo911: This is also defined in ion_stream.c but with some platform-specific logic, so this header definition is always overwritten.

ionc/ion_symbol_table_impl.h
```diff
@@ -105,7 +105,7 @@ iERR _ion_symbol_local_copy_same_owner(void *context, void *dst, void *src, int3
 iERR _ion_symbol_local_copy_new_owner(void *context, void *dst, void *src, int32_t data_size);
 iERR _ion_symbol_table_local_import_copy_new_owner(void *context, void *dst, void *src, int32_t data_size);
 iERR _ion_symbol_table_local_import_copy_same_owner(void *context, void *dst, void *src, int32_t data_size);
-
+iERR _ion_symbol_table_create_substitute(ION_SYMBOL_TABLE_IMPORT* import, ION_CATALOG* catalog, ION_SYMBOL_TABLE** result);
```
mijamo911: Looks like I forget to include the declaration :/

test/ion_assert.h
```diff
@@ -1,4 +1,5 @@
 #include <ion_debug.h>
+#include <ion_helpers.h>
```
mijamo911: Needed for the utility for shortening filenames

test/ion_binary_test.c
```diff
@@ -25,16 +25,16 @@ iERR test_ion_binary_len_uint_64() {
...
-    ASSERT_EQUALS_INT(2, ion_binary_len_int64(-255LL), "Wrong int64 length");
-    ASSERT_EQUALS_INT(2, ion_binary_len_int64(256LL), "Wrong int64 length");
-    ASSERT_EQUALS_INT(2, ion_binary_len_int64(-256LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(0, ion_binary_len_int_64(0LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(1, ion_binary_len_int_64(1LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(1, ion_binary_len_int_64(-1LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(2, ion_binary_len_int_64(255LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(2, ion_binary_len_int_64(-255LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(2, ion_binary_len_int_64(256LL), "Wrong int64 length");
+    ASSERT_EQUALS_INT(2, ion_binary_len_int_64(-256LL), "Wrong int64 length");
```
mijamo911: Renamed since the underlying method was renamed

test/tester.h
```diff
@@ -167,3 +167,9 @@ iERR test_reader_one_file_file_only             (char *filename);
 iERR test_reader_read_all_just_file             (hREADER hreader);
 iERR test_reader_read_all_just_file_byte_by_byte(hREADER hreader);
 iERR test_stream_handler                        (struct _ion_user_stream *pstream);
+
+iERR test_step_out_nested_s_expressions();
```
mijamo911: More missing declarations

tools/ionsymbols/ionsymbols.c
```diff
@@ -21,12 +21,14 @@
 
 #define IZ_INITIALIZE
 #include "ionsymbols.h"
-#include "options.h"
+
+#include <assert.h>
+#include <ion_helpers.h>
```
mijamo911: Sorted and added some missing includes

ionc/ion_reader_binary.c
tgregg: Hmm... based on your comment this looks like it should have been completely broken before. No test coverage?

tgregg:
Looks fine.

ionc/ion_reader_binary.c
mijamo911: Probably not, but I'm working on it...

PR #33, opened 2017-04-20
Fixes broken ion stream flushing logic; correctly names symbol writin…

tgregg:
…g method to match its header.

The removed code appears to have been copied/pasted incorrectly (as evidenced by the duplication and weird indentation/brace placement).

The change in ion_writer makes the implementation of the method match the name in the header. So that it has the potential to be used.


ionc/ion_writer.c
```diff
@@ -1048,7 +1048,7 @@ iERR _ion_writer_write_timestamp_helper(ION_WRITER *pwriter, ION_TIMESTAMP *valu
     iRETURN;
 }
 
-iERR ion_writer_write_symbol_id(hWRITER hwriter, SID value)
+iERR ion_writer_write_symbol_sid(hWRITER hwriter, SID value)
```
therapon: Seems that the function is not used. Also should the helper also get renamed?

tgregg: It's not used internally, but it is a public API. Sure, the helper function can change names too.

therapon: Nevermind found it.

pbcornell:
LGTM

PR #34, opened 2017-04-20
Fixes ion_int_compare, which was never setting its output parameter.

tgregg:
Pretty simple bugs.


ionc/ion_int.c
```diff
@@ -130,51 +129,52 @@ iERR ion_int_compare(ION_INT *iint1, ION_INT *iint2, int *p_result)
...
-        diff = 0;
+        *p_result = 0;
         SUCCEED();
     }
 
     IONCHECK(ion_int_is_null(iint1, &is_null1));
     IONCHECK(ion_int_is_null(iint2, &is_null2));
     if (is_null1 || is_null2) {
-        diff = (is_null1 - is_null2);  // TODO : really?
+        *p_result = (is_null1 - is_null2);  // TODO : really?
```
tgregg: I can only speculate, which probably means it should be removed. The behavior is not incorrect.

pbcornell:
Thanks for fixing an infinite loop, too!  LGTM

PR #35, opened 2017-04-20
Adds API for decimal equivalence under the Ion data model.

tgregg:
See the spec for the intricacies of decimal equivalence.


ionc/ion_decimal.c
```diff
@@ -0,0 +1,54 @@
...
+            != decQuadGetCoefficient(&right_canonical, right_coefficient)) {
+        goto not_equal;
+    }
+    if (memcmp(left_coefficient, right_coefficient, DECQUAD_Pmax)) {
+        goto not_equal;
+    }
+    *is_equal = TRUE;
+    SUCCEED();
+not_equal:
+    *is_equal = FALSE;
```
pbcornell: Should we have a `SUCCEED()` call here, too?  (using `ion_timestamp_equals` for comparison)

tgregg: SUCCEED is a macro that just sets the error code to IERR_OK (which it must already be, since we're still executing within this function) and jumps to the `done` label (which is contained in the iRETURN macro). There's no need to SUCCEED when you're unconditionally heading to iRETURN. SUCCEED is sometimes used redundantly like this in existing parts of the code base, but I don't plan on continuing that.

pbcornell: Sounds fine;  may cause some minor confusion, but seems reasonable.

PR #36, opened 2017-04-20
Adds API for comparing timestamps for instant equality.

tgregg:
ion-tests has two different timestamp equality semantics: equality under the data model (see the spec), and "instant" equality. The latter measures whether two timestamps represent the same instant in time without caring about precision or having the same local offset.

The `_ion_timestamp_to_utc` method has been added because the `ion_timestamp_to_time_t` has platform-specific dependencies and can only handle dates representable in `time_t`.


ionc/ion_timestamp.c
```diff
@@ -930,26 +1031,42 @@ iERR ion_timestamp_equals(const ION_TIMESTAMP *ptime1, const ION_TIMESTAMP *ptim
...
-    IONCHECK(ion_timestamp_to_time_t(ptime1, &t1));
-    IONCHECK(ion_timestamp_to_time_t(ptime2, &t2));
-    if (t1 != t2) {
-        goto is_false;
+    if (instant_only) {
+        // Fractional precision does not matter in this case. Trim any trailing zeros from the fractions, then compare.
+        decQuadReduce(&fraction_trimmed1, &ptime1->fraction, pcontext);
+        decQuadReduce(&fraction_trimmed2, &ptime2->fraction, pcontext);
+        ion_decimal_equals(&fraction_trimmed1, &fraction_trimmed2, pcontext, &decResult);
+        if (!decResult) goto is_false;
```
pbcornell: avoiding use of {} for simple if statements like this tends to cause more problems than it solves;  my preference is to always use {}

ionc/ion_timestamp.c
```diff
@@ -930,26 +1031,42 @@ iERR ion_timestamp_equals(const ION_TIMESTAMP *ptime1, const ION_TIMESTAMP *ptim
...
+            goto is_false;
         }
-        else {
+
+        if (HAS_TZ_OFFSET(ptime1) ^ HAS_TZ_OFFSET(ptime2)) {
             goto is_false;
         }
+        if (precision1 == ION_TS_FRAC) { // Then precision2 is also ION_TS_FRAC.
+            ion_decimal_equals(&(ptime1->fraction), &(ptime2->fraction), pcontext, &decResult);
+            if (!decResult) goto is_false;
```
pbcornell: ditto -- i'd suggest always using {}

ionc/ion_timestamp.c
```diff
@@ -930,26 +1031,42 @@ iERR ion_timestamp_equals(const ION_TIMESTAMP *ptime1, const ION_TIMESTAMP *ptim
...
+        }
+        ptime1_compare = *ptime1;
+        ptime2_compare = *ptime2;
+    }
+
+    if    ((ptime1_compare.year != ptime2_compare.year)
```
pbcornell: I suspect all these nested parentheses aren't needed;  not a big deal either way

ionc/inc/ion_timestamp.h
```diff
@@ -119,6 +119,12 @@ ION_API_EXPORT iERR ion_timestamp_to_time_t(const ION_TIMESTAMP *ptime, time_t *
 ION_API_EXPORT iERR ion_timestamp_equals(const ION_TIMESTAMP *ptime1, const ION_TIMESTAMP *ptime2,
         BOOL *is_equal, decContext *pcontext);
 
+/** Compare timestamps for instant equality only (i.e. precision and local offsets need not be equivalent).
+ *  NOTE: if this has any use externally, it could be exposed. If not, it should be removed.
```
tgregg: Fixed in #57

ionc/ion_timestamp.c
```diff
@@ -916,12 +916,113 @@ BOOL _ion_timestamp_is_leap_year(int y)
...
+        pout->month += 1;
+        pout->day = 1;
+    }
+    else {
+        SUCCEED();
+    }
+
+    if (pout->month < 1) {
+        pout->year -= 1;
+        pout->month = 11;
```
tgregg: Fixed in #57 . ion-tests coverage added in https://github.com/amzn/ion-tests/pull/28

PR #38, opened 2017-04-20
Adds support for symbol 0; implements proper unknown max ID semantics…

tgregg:
… for symbol table imports; disallows ordered structs with zero elements.

The spec now allows the symbol 0 (see the spec for the significance of that).

For shared symbol table imports, when the max ID of the import is null, not an int, or less than zero, it is treated as if it is undefined. When it is undefined and no exact match is found in the catalog, an error must be raised. See https://amznlabs.github.io/ion-docs/symbols.html .

There is a special type ID (0xD1) in Ion binary for "ordered structs" (elements sorted by field name symbol ID). The spec requires these not to be empty.


pbcornell:
lgtm

PR #39, opened 2017-04-20
Adds API to detect negative-zero floats.

tgregg:
Negative-zero is considered distinct from positive-zero when determining Ion float equivalence under the data model. This provides an API for determining if a float (double) is negative-zero.


pbcornell:
ship it!

PR #43, opened 2017-04-20
Adds support for reading text ints in the binary radix.

tgregg:
In addition to the hex radix, Ion ints may be expressed in the binary radix in the text format.


ionc/ion_reader_text.c
```diff
@@ -1432,12 +1445,20 @@ iERR _ion_reader_text_read_int64(ION_READER *preader, int64_t *p_value)
         magnitude = STR_TO_UINT64(value_start + 1, &value_end, 10);
     }
     else if (text->_value_sub_type == IST_INT_POS_HEX) {
-        // base is now 16, and we add 3 to the start for the "0x"
-        magnitude = STR_TO_UINT64(value_start + 2, &value_end, 16);
+        // base is now 16, and we add 2 to the start for the "0x"
```
pbcornell: thanks for correcting this

PR #45, opened 2017-04-20
Fixes text serialization of timestamps with large zero fractional pre…

tgregg:
…cision; ignores timestamp fractions with coefficient of zero and exponent >= 0, in accordance with the spec.

`decQuadToString` expresses highly-precise forms of zero using scientific notation. That was previously not supported when writing the timestamp as text Ion.

Additionally, the spec requires timestamp fractions of zero with >=0 exponent to be ignored (note that there's no way to represent this in text--only binary).


ionc/ion_timestamp.c
```diff
@@ -224,6 +224,23 @@ iERR ion_timestamp_to_string(ION_TIMESTAMP *ptime, char *buffer, SIZE buf_length
         if ((*cp) == '.') {
             IONCHECK(_ion_timestamp_copy_to_buf(pos, cp, end_of_buffer, &count));
         }
+        else if ((*cp) == 'E' || (*cp) == 'e') {
+            // This fraction is of the form 0E-N
+            IONCHECK(_ion_timestamp_copy_to_buf(pos, ".", end_of_buffer, &count));
+            pos += count;
+            if (pos >= end_of_buffer) FAILWITH(IERR_BUFFER_TOO_SMALL);
+            for (i = 0; i < -decQuadGetExponent(&ptime->fraction); i++) {
```
pbcornell: `i` is only needed for this loop;  consider declaring it here

PR #54, opened 2017-05-02
Fixes bugs that prevented writing of binary -0 floats and limited precision when writing text floats.

tgregg:
Note that "%20g" refers to the width of the string representation, whereas "%.20g" means that 20 digits of precision after the decimal point is required. It's unclear what the magic number should be, but this one passes all of our tests. This is probably not the best solution, but it is the least complex one (probably by far).


ionc/ion_binary.c
```diff
@@ -117,7 +117,7 @@ int ion_binary_len_ion_decimal(decQuad *value, decContext *context )
 int ion_binary_len_ion_float( double value )
 {
     int len = 0;
-    if (value == 0) {
+    if (value == 0 && !ion_float_is_negative_zero(value)) {
```
pbcornell: len is already initialized to 0 -- there's no need to set it to 0 again in this if block.  Consider simplifying to:
```
int len = 0;
if (value != 0 || ion_float_is_negative_zero(value)) {
    len = sizeof(value);
}
return len;
```

PR #58, opened 2017-05-11
Fixes a symmetrical bug in which the binary reader and writer did not…

tgregg:
… convert timestamps from/to UTC on de/serialization.

This is (unfortunately) required by the spec. The bug did not surface in roundtrips because it was completely symmetrical in the reader and writer. Added a test case.

### DIFF
--- a/ionc/ion_timestamp.c
+++ b/ionc/ion_timestamp.c
@@ -951,11 +951,12 @@ iERR _ion_timestamp_to_utc(const ION_TIMESTAMP *ptime, ION_TIMESTAMP *pout) {
 
     int8_t sign;
     uint16_t  min_shift, hour_shift = 0;
+    int16_t offset;
     BOOL needs_day = FALSE;
 
     if (!ptime || !pout) FAILWITH(IERR_INVALID_ARG);
 
-    _ion_timestamp_initialize(pout);
+    offset = ptime->tz_offset;
     pout->year = ptime->year;
     pout->month = ptime->month;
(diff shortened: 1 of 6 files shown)


ionc/ion_timestamp.c
```diff
@@ -1305,6 +1308,9 @@ iERR ion_timestamp_binary_read(ION_STREAM *stream, int32_t len, decContext *cont
     // now we set the "has timezone" bit
     if (has_offset) {
         SET_FLAG_ON(ptime->precision, ION_TT_BIT_TZ);
+        ptime->tz_offset = (int16_t)-offset;
+        IONCHECK(_ion_timestamp_to_utc(ptime, ptime));
+        ptime->tz_offset = (int16_t)offset;
```
tgregg: Because of how timestamps are stored in Ion binary. All of their fields are stored in UTC, but the offset field still refers to the local offset from UTC. So when binary timestamps are read, they need to be converted from UTC to local time, with the offset preserved. The simplest way to do this is to negate the offset, perform a UTC conversion on the resulting timestamp, and then restore the correct offset. That's what this does.

pbcornell: ic, thanks for explaining.  i do wonder whether we could remove line 1215, as it doesn't appear to be adding any value (tho maybe i'm missing something...).

tgregg: Line 1215 is removed in #59 .

tgregg:
Merged to master.

PR #59, opened 2017-05-12
Fixes a bug that caused superfluous timestamp local offsets to be read and written for timestamps with less than minute precision.

tgregg:
See https://github.com/amzn/ion-tests/pull/29 for more info. In addition to the test cases added here, ion-c passes the test vectors added in the linked ion-tests pull request.


ionc/ion_timestamp.c
```diff
@@ -1307,6 +1309,7 @@ iERR ion_timestamp_binary_read(ION_STREAM *stream, int32_t len, decContext *cont
 timestamp_is_finished:
     // now we set the "has timezone" bit
     if (has_offset) {
+        ptime->tz_offset = offset;
```
tgregg: Nope, that's superfluous. Not sure how it got there -- maybe leftover from a merge or something. Will remove.

pbcornell: ok, thanks;  otherwise looks good, ship it!

tgregg:
Merged to master.

PR #60, opened 2017-06-16
Consolidates and improves symbol handling.

tgregg:
* Fixes symbol processing to never add system semantics to quoted symbol identifiers.
  * In Ion text, symbol identifiers have the form `$<int>`, and represent a SID mapping in the current symbol table context. When a quoted symbol token with the same form as a symbol identifier is encountered, it should be not be treated as a SID and should not have its text transformed. When the user uses a writer API to write a symbol from a given string and provides a string with the form of a symbol identifier, this symbol token should be written as a quoted symbol token so that readers do not treat this symbol with special semantics. If a user wants to write a symbol token with unknown text in range of the current symbol table context, he/she can use an API that accepts an integer symbol ID. If that ID maps to a symbol with unknown text, text writers will write an unquoted symbol identifier.
* Fixes a bug that caused top-level unannotated IVM-esque symbols not to be treated as a no-ops on read and write.
  * According to the spec, there are several top-level symbol values that look like the IVM, but are treated as a no-op, including:
    * Any symbol identifier that maps to a symbol with the text `$ion_1_0`, including `$2`.
    * `'$ion_1_0'`, a quoted symbol with the same text as the IVM.
* Consolidates all symbol table processing logic in ion_symbol_table.c. Previously, much of this logic was needlessly duplicated in the binary reader (~400 lines deleted).
* Fixes a bug that prevented correct handling of not-found shared symbol table imports; uses delegate (non-flattened) symbol table lookups to avoid allocating NULL slots for not-found imports.
  * Previously, entering a new symbol table context involved copying all symbols from the system table and shared tables into a flattened collection within the local table, and appending any local symbol tables encountered later. This technique, which is costly upon switching symbol table contexts (especially for system symbols, which don't change (yet)), failed to handle not-found shared imports correctly, and would have required allocating NULL slots for all not-found symbols. Using delegate lookups (where the local symbol table simply refers to the system table and any shared tables, and queries them in the correct order) solves these problems.
* Removes the need for the text reader to query/create a local symbol table for symbols with known text; adds internal APIs to read symbol values as ION_SYMBOL, which enables correct roundtripping of symbols with unknown text.
  * Previously, encountering a symbol token in Ion text involved interning that symbol into a local symbol table context. This is only actually required for symbol identifiers, which don't represent the literal symbol text. All other text symbols have their text represented verbatim in the encoding, and do not require costly lookups or interning.
  * In order to correctly roundtrip a symbol with unknown text, its symbol ID is needed. Existing APIs enabled reading of the current symbol value's SID or text, but not both. Calling both would cause invalid state because the readers expect only one call to a `ion_reader_read*` API per call to `ion_reader_next`. The new function, which returns both in the form of an ION_SYMBOL, remains internal for now (it is used by `ion_writer_write_all_values`, which roundtrips all values from the given reader). We can make it public at a later date if necessary.

### DIFF
--- a/ionc/inc/ion_reader.h
+++ b/ionc/inc/ion_reader.h
@@ -300,6 +300,7 @@ ION_API_EXPORT iERR ion_reader_get_type            (hREADER hreader, ION_TYPE *p
 ION_API_EXPORT iERR ion_reader_has_any_annotations (hREADER hreader, BOOL *p_has_any_annotations);
 ION_API_EXPORT iERR ion_reader_has_annotation      (hREADER hreader, iSTRING annotation, BOOL *p_annotation_found);
 ION_API_EXPORT iERR ion_reader_is_null             (hREADER hreader, BOOL *p_is_null);
+ION_API_EXPORT iERR ion_reader_is_in_struct        (hREADER preader, BOOL *p_is_in_struct);
 ION_API_EXPORT iERR ion_reader_get_field_name      (hREADER hreader, iSTRING p_str);
 ION_API_EXPORT iERR ion_reader_get_field_sid       (hREADER hreader, SID *p_sid);
 ION_API_EXPORT iERR ion_reader_get_annotations     (hREADER hreader, iSTRING p_strs, SIZE max_count, SIZE *p_count);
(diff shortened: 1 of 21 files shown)


tgregg:
Merged to master. Leaving open for a week in case there is any feedback that can be iterated upon.

PR #62, opened 2017-07-29
Implements fully arbitrary decimal precision through new ION_DECIMAL APIs.

tgregg:
Previously, the public Ion decimal APIs exposed decNumber's `decQuad` module directly. Because `decQuad` is limited to 34 digits of precision, this meant that ion-c's decimal implementation could not be fully compliant with the Ion specification, which requires arbitrary-precision decimals.

The decNumber library provides another module (confusingly named `decNumber`) which does support arbitrary precision essentially up to the memory constraints of the platform. Unfortunately, the performance comparisons between `decQuad` and `decNumber` (included in the decNumber documentation) show that `decNumber` is often several times slower than `decQuad`.

To avoid significant performance regressions for decimal numbers that can fit in a `decQuad` (i.e. the most commonly-used decimal numbers), it is necessary to be able to switch between the `decQuad` and `decNumber` representations, using `decQuad` when possible and `decNumber` only when necessary. The new `ION_DECIMAL` type abstracts this behavior from the user, while providing the ability for ion-c to change its backing decimal library, if necessary, without impacting the public APIs.

Under the hood, the new `ION_DECIMAL` computational APIs are simply shims to the decNumber library. There is some logic on top that reconciles operands that are backed by different types and handles overflow from `decQuad` to `decNumber`. The user does not need to know or care when this happens.

The existing `decQuad` APIs have been deprecated, but not removed. In order to support reading and writing decimals with more than 34 digits of precision, the user will need to provide a `decContext` with a higher limit in their `ION_READER_OPTIONS` AND `ION_WRITER_OPTIONS`. No compile-time configuration is required.

In addition to passing the existing decimal vectors in ion-tests, `test_ion_decimal` has been added.


tgregg:

PR #65, opened 2017-08-08
Implements read support for appending symbols to the current local symbol table.

tgregg:
This is achieved by cloning the previous local symbol table (if necessary) and appending new symbols to the clone.

Also makes a few improvements around symbol table handling:
* Consolidates the text and binary readers' potential symbol table processing logic into `_ion_reader_process_possible_symbol_table`. Previously, the two readers used different logic, which led to the binary reader not supporting adding shared symbol tables to the catalog and using the incorrect local symbol table owning pool.
* In accordance with the spec, any value in the symbols list that is null or is not a string is treated as a valid SID mapping with unknown text.
* A symbol table struct must not declare more than one `symbols` or `imports` list. Allowing this could lead to inconsistent mappings since struct fields are unordered.

Testing: passes all existing tests as well as the new LST-append and null slot vectors added in https://github.com/amzn/ion-tests/pull/35 . Once that PR is merged, ion-c's ion-tests submodule will be pointed to the latest revision.


tgregg:

PR #66, opened 2017-08-08
Checks only the first annotation on a potential symbol table struct.

tgregg:
Local symbol table structs may have additional annotations as long as the $ion_symbol_table annotation comes first. See the very bottom of http://amzn.github.io/ion-docs/symbols.html .

https://github.com/amzn/ion-tests/pull/36 adds ion-tests coverage for this. ion-c's ion-tests submodule will be updated once that PR is merged.


tgregg:

Issue #68, opened 2017-08-21
Configure Travis CI to build and run tests


tgregg:

tgregg:
#82 adds gcc/clang on MacOS/Ubuntu.

Consider adding a matrix that tests multiple compiler versions on each OS.

PR #70, opened 2017-10-06
Avoids switching on intptr_t in C++ tests. Aligns ION_TYPE and TID constants.

tgregg:
Successful build/test runs:
* MacOS 10.11.6 with clang-800.0.42.1
* AL2012.03 with gcc 4.4.6 20110731

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ test/all_tests
 tools/ionizer/ionizer
 tools/ionsymbols/ionsymbols
 build/
+cmake-build-debug/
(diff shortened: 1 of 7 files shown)


tgregg:
Note: not sure why Travis CI checks were run. We haven't configured Travis CI, and this change didn't add any additional configuration. It looks like a default Travis CI configuration was used, which fails when it doesn't find a rakefile.

#68 was created to add proper Travis CI support for running the tests, and should be prioritized.

therapon:
LGTM

PR #71, opened 2017-10-19
Adds support for writing local symbol tables using special append syntax.

tgregg:
The major benefit of LST-append is that it allows writers to flush their streams while retaining the same symbol table context and without having to rewrite the entire context.

This provides a way to do that naturally through the writer APIs. Basically, if the user `flush`es a writer, the stream will be flushed, but symbol table context will be retained. The writer will note the `max_id` of the symbol table when it was flushed so that on the next flush, it can append just the symbols that occurred after that point.

If the user `finish`es a writer, the stream will be flushed, and the symbol table context will be reset (as before).


tgregg:

PR #73, opened 2017-10-19
Forces the writer to raise an error if the user attempts to write an out-of-range symbol ID.

tgregg:
This prevents the writer from writing Ion data that can never successfully be read; data with symbol IDs out of range of the current local symbol table context MUST raise an error on read.


tgregg:

PR #74, opened 2017-10-19
Writes a text LST when the writer has imports; ensures the import declaration's max_id is authoritative; prohibits writer flush/finish/close when a value is in progress.

tgregg:
In order to avoid having to buffer text streams, when text writers are provided shared symbol tables, a local symbol table that declares those imports will always be written (before the first value is written).

The max_id of the import declaration must be authoritative; this helps allow forward-compatibility when symbols are appended to a given shared symbol table.

Flush/finish must happen between top-level values because changes to the local symbol table context can only occur at the top level.


tgregg:

PR #75, opened 2017-10-19
Makes writer options' imports list an ION_COLLECTION of ION_SYMBOL_TABLE_IMPORT to support writing with not-found imports.

tgregg:
The writer should be able to write symbol tokens that map to shared symbol table imports even when those shared symbol table imports can't be resolved. A reader that later consumes the data will be able to resolve the symbol tables to known text if it can resolve the import declarations.


tgregg:

PR #76, opened 2017-10-19
 Adds support for intercepting user-written local symbol tables.

tgregg:
When a user manually writes a top-level struct with the $ion_symbol_table annotation, it must be intercepted and treated as a local symbol table. This avoids the writer blindly writing what would look to a reader like a symbol table context change, without the writer having internally changed its symbol table context.


tgregg:

PR #77, opened 2017-10-19
 Adds import location to ION_SYMBOL to fully support roundtripping symbols with unknown text; exposes public APIs for reading and writing ION_SYMBOLs.

tgregg:
Import location, defined as an import name and position within that import, is essential for determining equivalence of symbol tokens with unknown text. Symbol tokens with unknown text but equivalent import locations are defined as equivalent by the spec. This also allows a writer to correctly write symbol tokens with unknown text.


tgregg:

PR #78, opened 2017-10-19
Adds ability to register a callback with the reader to be notified upon change of local symbol table context.

tgregg:
It is a common pattern to read Ion data, operate on a subset of the values in the stream, and re-write it so that the next node has a full view of the data.

Because Ion streams may have multiple symbol table contexts, if the reader doesn't convey when the context has changed, the writer won't be able to synchronize its symbol table context accordingly, which can lead to errors or incorrect symbol mappings on write.

This change allows the user to be notified when a reader has encountered a change in the symbol table context that changes which shared symbol table imports are required to interpret the symbol tokens. The user may then pass these along to the writer, which will flush its previous symbol table context and create a fresh one which uses the provided shared symbol tables. This is also done automatically by `ion_writer_write_all_values`.

### DIFF
--- a/ionc/inc/ion_reader.h
+++ b/ionc/inc/ion_reader.h
@@ -22,6 +22,34 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/**
+ * A function that may be called by the reader upon a change to the stream's symbol table context.
+ *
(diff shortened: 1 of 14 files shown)


tgregg:

PR #79, opened 2017-10-19
Adds support for handing off shared symbol tables from reader to writer in roundtrip tests.

tgregg:
This allows the tests to correctly roundtrip valid Ion data with not-found imports, and leads to the removal of `good/item1.10n` from the skip list.


tgregg:

PR #80, opened 2017-10-19
Treats all local symbol tokens with unknown text as symbol zero.

tgregg:
Because local symbol tables are included in the data and may not change, a local symbol token with unknown text (due to a null slot in the local symbol table's `symbols` list) can NEVER have known text (unlike a shared symbol token with unknown text). Therefore, treating all such symbol tokens as symbol zero does not result in data loss. Furthermore, it simplifies the implementation because roundtripping such symbol tokens does not require inserting null slots into the writer's local symbol table; symbol zero is implicitly present in all local symbol table contexts.


tgregg:

PR #81, opened 2017-10-19
Removes public APIs that allow directly reading or writing local SIDs.

tgregg:
Such APIs are problematic because the symbol table context of the writer is not guaranteed to be the same as that of the reader; the same local SID can map to different text in both.

APIs which use ION_SYMBOLs, which convey import location, should be used instead.


tgregg:

PR #92, opened 2017-11-25
Merges all outstanding pull requests.

tgregg:


tgregg:
Succeeds on Windows: https://ci.appveyor.com/project/tgregg/ion-c/build/1.0.9
Succeeds on Mac, but a few tests fail on Linux: https://travis-ci.org/amzn/ion-c/builds/306979937?utm_source=github_status&utm_medium=notification

Investigating.

tgregg:
A note on the fixes:

1. Casting an `int32_t *` to a `int_fast32_t *` is not safe, because `int_fast32_t` MAY be larger than 32 bits.
2. Different compiler implementations appear to handle the order of operations in `a |= (a &= c)` differently. It is safer to split this into multiple expressions so that the intent is clear.

PR #100, opened 2018-03-13
Fixes bitwise calculations for extractor path bitmaps when the bitmaps span multiple integer units.

tgregg:
*Description of changes:*

Fixes a few bugs in the bitwise shifting and comparison logic when multiple integers are required to hold a single path bitmap. These include:

* Improper use of sizeof() to calculate the size of an array provided as a function parameter. This was fixed by calculating the size of the (fixed-size) array manually.
* Missing parentheses around preprocessor parameters that could change the intended meaning.
* Left-shifting unadorned `1`, which has undesired results for shifts greater than 31 places.

### DIFF
--- a/ionc/ion_extractor.c
+++ b/ionc/ion_extractor.c
@@ -16,23 +16,25 @@
 #include "ion_extractor_impl.h"
 
 #if ION_EXTRACTOR_MAX_NUM_PATHS > ION_EXTRACTOR_MAX_NUM_PATHS_THRESHOLD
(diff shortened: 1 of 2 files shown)


tgregg:
Regarding testing:
Testing this requires defining a macro above its default. We need coverage both at and above the default. Typically users choose a value and add its definition as a compiler flag. I tested this manually to verify the fix works, but we don't currently have an automated way of doing this that lets us test multiple values for the macro.

dlurton:
Since this has now at least once given an issue, does it make sense to at least make an issue for investigating how to test different values for the macro in an automated fashion?

tgregg:
@dlurton Yes it does. See #101 .

PR #108, opened 2018-06-22
Fixes an array-out-of-bounds bug in ion_extractor_path_create_from_ion.

tgregg:
*Description of changes:*

When creating a path from Ion data, the length needs to be checked after each path component is iterated to verify that the max length isn't exceeded. Failing to do this can cause the pointer to go past the end of the array when attempting to create a path that exceeds `ION_EXTRACTOR_MAX_PATH_LENGTH` using this function.

Successful AppVeyor build: https://ci.appveyor.com/project/tgregg/ion-c/build/1.0.30

### DIFF
--- a/ionc/ion_extractor.c
+++ b/ionc/ion_extractor.c
@@ -204,6 +204,8 @@ iERR ion_extractor_path_create_from_ion(ION_EXTRACTOR *extractor, ION_EXTRACTOR_
     ASSERT(callback);
     ASSERT(ion_data_length > 0);
     ASSERT(p_path);
+    ASSERT(extractor->_options.max_path_length <= ION_EXTRACTOR_MAX_PATH_LENGTH);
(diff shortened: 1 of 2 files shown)


ionc/ion_extractor.c
```diff
@@ -220,6 +222,9 @@ iERR ion_extractor_path_create_from_ion(ION_EXTRACTOR *extractor, ION_EXTRACTOR_
     }
     IONCHECK(ion_reader_step_in(reader));
     for (;;) {
+        if (path_length >= extractor->_options.max_path_length) {
+            FAILWITHMSG(IERR_INVALID_ARG, "Illegal number of path components.");
```
almann: I might make the message indicate the number given and the maximum exceeded to be a bit more informative.

dlurton: Yes, or at least change it to say "Maximum number of path components exceeded" or somesuch.

PR #110, opened 2018-07-06
Inverts tid_*_INT and tid_* declarations to enable switch statements on newer C++ compilers.

tgregg:
*Description of changes:*
Many C++ compilers don't allow casts to pointers in switch blocks, causing failures on macros like tid_INT_INT. [This](https://github.com/amzn/ion-c/commit/f6811b256465cca8572bc37b7f923762e0cabe66) commit shows where this change was necessary in the ion-c tests, which require a C++ compiler. This change makes the new macros available publicly and uses them within the tools and test modules.

### DIFF
--- a/ionc/inc/ion_types.h
+++ b/ionc/inc/ion_types.h
@@ -70,44 +70,44 @@ typedef struct ion_type    *ION_TYPE;
 /** Current value type not known (has not been checked)
  *
  */
-#define tid_none      ((ION_TYPE)(-0x200))
+#define tid_none      ((ION_TYPE)(tid_none_INT))
 
 /** Stands for both End of File and End of Container, unfortunately.
  */
-#define tid_EOF       ((ION_TYPE)(-0x100))
+#define tid_EOF       ((ION_TYPE)(tid_EOF_INT))
 
 /** Ion Value Type NULL.*/
-#define tid_NULL         ((ION_TYPE) 0x000)
-#define tid_BOOL         ((ION_TYPE) 0x100)
-#define tid_INT          ((ION_TYPE) 0x200)
-#define tid_FLOAT        ((ION_TYPE) 0x400)
-#define tid_DECIMAL      ((ION_TYPE) 0x500)
-#define tid_TIMESTAMP    ((ION_TYPE) 0x600)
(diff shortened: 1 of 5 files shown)


ionc/inc/ion_types.h
```diff
@@ -109,6 +109,29 @@ typedef struct ion_type    *ION_TYPE;
 #define tid_SEXP_INT       ION_TYPE_INT(tid_SEXP)
 #define tid_DATAGRAM_INT   ION_TYPE_INT(tid_DATAGRAM)
```
almann: Would inverting the definition work for this instead (without adding the `_TYPE` suffix macros?

E.g.
```c
#define tid_NULL_INT   0x000
...
#define tid_NULL  ((ION_TYPE) tid_NULL_INT)
```

This has the benefit of not introducing a separate set of type integers, but I could be missing something more to this change.

pbcornell:
Would prefer we avoid what is effectively duplication between the tid_* and TYPE_* defs, but sounds like newer C++ compilers are forcing the issue.  Beyond that, my only concern is naming.  We currently have:

ION_TYPE_INT() : intptr_t
tid_BOOL       : ION_TYPE
tid_BOOL_INT   : intptr_t

and the proposed changes add:

ION_TID_INT()   : int
TYPE_BOOL       : int

Note the asymmetry here:  ION_TYPE_INT() leverages tid_* constants, whereas ION_TID_INT() leverages TYPE_* constants;  ideally, imho, ION_TYPE_INT() would use TYPE_* constants, and ION_TID_INT() would use TID_* constants.  Part of that ship has sailed, but it makes me wonder if the new constants should be named TID_BOOL, for example (or perhaps ION_TID_BOOL)?  And then there's likely to be confusion around tid_BOOL, tid_BOOL_INT, and TID_BOOL/ION_TID_BOOL....

Thoughts?

ionc/inc/ion_types.h
tgregg: Yes, that should work.

tgregg:
@pbcornell I agree. The problem is we already have TID_* constants defined in the internal ion_const.h header. Let me try Almann's suggestion and see if that cleans this up.

Issue #132, opened 2019-04-02
Public Header Structure Should Match Install Target Headers


almann:
Currently when using he `install` target, public headers get placed into an `ionc` directory.  However within the project this is just is defined as flat in the `ionc/inc` directory.

Within the package we will include a public header such as `ion.h` as the following forms:

* `#include "ion.h"`
* `#include <inc/ion.h>"`

If one uses CMake to nest ion-c directly, public headers point to the `ionc/inc` drectory, thus making the top-levle form correct `#include <ion_types.h>`.

If you use the installation target and consume ion-c as a package, this becomes `#include <ionc/ion_types.h>`.

This should be made consistent as it shouldn't matter if someone nests ion-c into their CMake project or consume it as an installed package.

This recently caused me some confusion when I ported some code assuming the installed package form to a nested CMake dependency.

PR #135, opened 2019-07-17
Refactor CMake file and position of header files

therapon:
*Description of changes:*

This PR is to obtain early feedback on the approach taken to update, refactor and address #132 . 
If this is ok, then there will be another commit to address CMake packaging as well as generation of .deb and .rpm packages. 


tools/events/CMakeLists.txt
```diff
@@ -1,3 +1,18 @@
...
-add_library(ion_events ion_event_util.cpp ion_event_stream.cpp ion_event_equivalence.cpp)
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+add_library(ion_events
+        ion_event_util.cpp
+        ion_event_stream.cpp
+        ion_event_equivalence.cpp)
+
+target_include_directories(ion_events
+        PUBLIC
+            ${CMAKE_CURRENT_SOURCE_DIR}/inc
```
pbcornell: should this be "include"?

decNumber/include/decNumber/decSingle.h
```diff
@@ -38,9 +38,9 @@
   #define DECSINGLE_Ehigh (DECSINGLE_Emax + DECSINGLE_Bias - (DECSINGLE_Pmax-1))
 
   /* Required includes                                                */
-  #include "decContext.h"
-  #include "decQuad.h"
-  #include "decDouble.h"
+  #include "include/decNumber/decContext.h"
+  #include "include/decNumber/decQuad.h"
+  #include "include/decNumber/decDouble.h"
```
pbcornell: change these 3 to use the `#include <decNumber/dec*.h>` form?

decNumber/include/decNumber/decDouble.h
```diff
@@ -38,7 +38,7 @@
   #define DECDOUBLE_Ehigh (DECDOUBLE_Emax + DECDOUBLE_Bias - (DECDOUBLE_Pmax-1))
 
   /* Required includes                                                */
-  #include "decContext.h"
+  #include "include/decNumber/decContext.h"
```
pbcornell: any reason not to use the following form here?

`#include <decNumber/decContext.h>`

decNumber/include/decNumber/decSingle.h
therapon: Yes. I thought I fixed these, apparently I missed a couple.

tools/events/CMakeLists.txt
therapon: This points to the subfolder for `event` which is only used internally in our code. I am not 100% sure we want this lib exposed to customers.

decNumber/include/decNumber/decDouble.h
therapon: ditto

therapon:
> It's been quite a while since I've written/reviewed C, and I don't recall the expectations around include paths. That said, something seems off when I see `#include "include/..."`, which occurs a few times below.

The use of quote in includes causes the preprocessor to first search in the current folder of the file before looking in system locations (folders specified by -I). The <> use looks in system locations. 
The pattern, I believe, is to use <> for .h files that are public and "" for private to the applications code.

The `include/..` is a mistake and I will clear those out. 
> 
> The general pattern #include <_libname_/_file.h_> feels like a good approach. I assume the #include <_file.h_> form is used to include implementation logic, not APIs.
> 

Yes. The pattern is to minimize conflicts in client code by having <libname/..> prefix. 

> Regarding the move from "inc" to "include": will this break consumers?
> 

It can. It depends how IonC was consumed. The follow up PR that will address CMake packaging will make integration with IonC more modular for CMake client code (and somewhat for non-CMake client code) that will make all the build configuration (-I options) controlled by CMake.

tools/events/CMakeLists.txt
tgregg: Yes, let's not expose the stuff in `events` right now.

tgregg:
The approach looks fine to me. Any idea why it's failing on Windows?

therapon:
> The approach looks fine to me. Any idea why it's failing on Windows?

Not yet. Working on reproducing on my machine.

tgregg:
Why do the headers to in `ionc/include/ionc` and not just `ionc/include`?

therapon:
> Why do the headers to in `ionc/include/ionc` and not just `ionc/include`?

Mimics the Install location `/usr/local/include/ionc` and allows code within the package and code outside the package to use <ionc/include> consistently.

therapon:
> Why do the headers to in `ionc/include/ionc` and not just `ionc/include`?

Mimics the Install location `/usr/local/include/ionc` and allows code within the package and code outside the package to use <ionc/include> consistently.

PR #137, opened 2019-09-08
Public/Private header structure fix and CMake updates

therapon:
*Issue #, if available:* #132 

*Description of changes:*

* Moves header files and ensures public headers structure and build header structure are the same. 
* Updates CMake version 


ionc/ion_errors.c
```diff
@@ -25,9 +25,9 @@ const char *ion_error_to_str(iERR err)
         break;
 
     #define ERROR_CODE( name, val ) case name: s = #name; break;
-    #include "ion_error_codes.h"
+    #include "ionc/ion_error_codes.h"
 
-    default:
+        default:
```
pbcornell: not sure why we're changing the indent here;  i expect this to line up with `case IERR_NOT_IMPL:` above

therapon: Fixed

PR #145, opened 2019-11-14
Fixes handling of timezone offsets in ion_timestamp_to_time_t

### LINKED ISSUE #144: ion_timestamp_to_time_t handles timezone offsets incorrectly
In function ion_timestamp_to_time_t, the timezone offset is added to the current time instead of being substracted.
See:
[ion_timestamp.c#L99](https://github.com/amzn/ion-c/blob/9e9087430866f5b957801dd25cda81755e6222a4/ionc/ion_timestamp.c#L99)

```
iERR ion_timestamp_to_time_t(const ION_TIMESTAMP *ptime, time_t *time)
{
...

    tm.tm_year = ptime->year - 1900;   //  tm_year: year since 1900
    tm.tm_mon = ptime->month -1;       // tm_mon: month (0 – 11, 0 = January)
    tm.tm_mday = ptime->day;           // tm_mday: day of the month (1 – 31)
    tm.tm_hour = ptime->hours;         // tm_hour: hour (0 – 23)
    tm.tm_min = ptime->minutes;        // tm_min: minutes (0 – 59)
    tm.tm_sec = ptime->seconds;        // tm_sec: seconds (0 – 60, 60 = Leap second)

    if (HAS_TZ_OFFSET(ptime)) {
        tm.tm_min += (ptime->tz_offset);
    }
```

This leads to wrong results when converting timestamps containing a timezone offset, e.g.:

Initial timestamp: "2020-07-01T14:24:57.798+01:00"
Expected result: 2020-07-01 13:24:57.798
Actual result: 2020-07-01 15:24:57.798

pbcornell:
Resolves #144 

Also adds some unit tests for ion_timestamp.c, and moves the existing `IgnoresSuperfluousOffset` test to be with its new siblings in test_ion_timestamp.cpp.

### DIFF
--- a/test/test_ion_timestamp.cpp
+++ b/test/test_ion_timestamp.cpp
@@ -13,70 +13,69 @@
  * permissions and limitations under the License.
  */
 
+#include <tuple>
 #include "ion_assert.h"
 #include "ion_helpers.h"
 #include "ion_event_util.h"
 #include "ion_test_util.h"
 #include "ion_event_equivalence.h"
 
-struct timestamp_test {
-    const char *str;
+class IonTimestamp : public ::testing::TestWithParam< ::testing::tuple<std::string, int, time_t, std::string> > {
+public:
+    std::string str;
     int local_offset;
     time_t time;
-    const char *str_unknown_offset;
+    std::string str_unknown_offset;
+
+    virtual void SetUp() {
+        str = ::testing::get<0>(GetParam());
+        local_offset = ::testing::get<1>(GetParam());
+        time = ::testing::get<2>(GetParam());
+        str_unknown_offset = ::testing::get<3>(GetParam());
+    }
 };
 
-const int num_tests = 3;
-struct timestamp_test tests[num_tests] = {
-        {"2020-07-01T14:24:57-01:00", -60, 1593617097, "2020-07-01T15:24:57-00:00"},
-        {"2020-07-01T14:24:57+00:00",   0, 1593613497, "2020-07-01T14:24:57-00:00"},
-        {"2020-07-01T14:24:57+01:00",  60, 1593609897, "2020-07-01T13:24:57-00:00"},
-};
+INSTANTIATE_TEST_CASE_P(IonTimestamp, IonTimestamp, testing::Values(
+        std::tr1::make_tuple(std::string("2020-07-01T14:24:57-01:00"), -60, 1593617097, std::string("2020-07-01T15:24:57-00:00")),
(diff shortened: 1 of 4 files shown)


test/test_ion_timestamp.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+struct timestamp_test {
+    const char *str;
+    int local_offset;
+    time_t time;
+    const char *str_unknown_offset;
+};
+
+const int num_tests = 3;
+struct timestamp_test tests[num_tests] = {
```
tgregg: These tests are good candidates for the native parameterization provided by Google test via `TEST_P`. This will eliminate the `for` loops in each test and allow the different cases to be listed and run individually in the IDE. Documentation here: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-value-parameterized-tests

The simplest example in ion-c is in `test_ion_symbols.cpp`, which contains many [`TEST_P` tests](https://github.com/amzn/ion-c/blob/master/test/test_ion_symbol.cpp#L150) that parameterize on `BinaryAndTextTest`. These parameters are instantiated [here](https://github.com/amzn/ion-c/blob/master/test/test_ion_symbol.cpp#L23), defined in `ion_test_util.h` [here](https://github.com/amzn/ion-c/blob/master/test/ion_test_util.h#L31) and implemented in `ion_test_util.cpp` [here](https://github.com/amzn/ion-c/blob/master/test/ion_test_util.cpp#L20).

pbcornell: I haven't been able to parameterize these tests using a struct due to a compilation error I'm not smart enough to fix.  Adding logic to provide names for each parameterized iteration of a test prevents the tests from showing in the CLion test display, so I've removed that aspect from the latest commit.

I appreciate parameterized tests, but I'm not fully sold on applying the idea here.  It mostly appears to add complexity for what is [currently] a fairly simple set of tests.  As this set of tests grows in size/complexity, perhaps parameterization would add more value.  What do you think?

tgregg: I still prefer the parameterized versions. I don't think there's a reason to undo that work now.

I encountered the problem of CLion not playing nicely with the named test instances in test_vectors, too. There, I hid the verbose names behind an ifdef that is not defined when running in CLion, allowing for the names to be printed when built elsewhere (the command line, Travis/AppVeyor). I don't think that's a big enough value add to spend time on here, though; the biggest benefit of parameterized tests is being able to run the particular instance that failed individually, which is still possible even without nice names.

pbcornell: Ok, still parameterized but without names, and the build now succeeds.

PR #156, opened 2019-12-20
Updates ion-tests and fixes a bug that prevented the text writer from writing timestamps with high-precision fractional seconds.

tgregg:
*Description of changes:*
* Updates the ion-tests submodule to commit [2a1f3f4](https://github.com/amzn/ion-tests/commit/2a1f3f4f272256108e89f1c53bfae5b3a9ede812), which is current as of December 20, 2019.
* Fixes a bug uncovered by the addition of `good/timestampsLargeFractionalPrecision.ion` that prevented the text writer from writing timestamps with high-precision, low-magnitude fractional seconds (e.g. 1E-8). Previously, the implementation required on `decQuadToString` to create the string representation for these fractions, but did not handle the case when the result was in scientific notation. `decQuadToString` does not allow the caller to specify that the expanded notation should always be used, so the fix expands the scientific notation manually.
* Ensures that the binary and text writers have the same behavior when attempting to write a timestamp with a fraction outside the interval [0, 1) (that is, to throw `IERR_INVALID_TIMESTAMP`). This involved refactoring the bounds verification into `_ion_timestamp_validate_fraction` and calling it in the binary writer, which previously did not validate the fraction.

### DIFF
--- a/ionc/ion_timestamp.c
+++ b/ionc/ion_timestamp.c
@@ -134,13 +134,86 @@ iERR _ion_timestamp_initialize(ION_TIMESTAMP *ptime)
     iRETURN;
 }
 
+/**
+ * Writes the given string containing a fraction expressed in scientific notation in expanded form. For example,
+ * the string "1E-8" will be written as ".00000001".
+ * @param scientific_fraction a string representing a number in the range [0, 1).
+ * @param exponent the exponent of the number denoted by `scientific_fraction`.
+ * @param buffer the destination for the expanded fraction.
+ * @param end_of_buffer the end of `buffer`.
+ * @param p_copied the number of bytes copied to `buffer`.
+ * @return IERR_INVALID_TIMESTAMP if the given string is not in the range [0, 1), or IERR_BUFFER_TOO_SMALL if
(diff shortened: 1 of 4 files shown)


ionc/ion_timestamp.c
```diff
@@ -134,13 +134,86 @@ iERR _ion_timestamp_initialize(ION_TIMESTAMP *ptime)
     iRETURN;
 }
 
+/**
+ * Writes the given string containing a fraction expressed in scientific notation in expanded form. For example,
+ * the string "1E-8" will be written as ".00000001".
+ * @param scientific_fraction a string representing a number in the range [0, 1).
+ * @param exponent the exponent of the number denoted by `scientific_fraction`.
+ * @param buffer the destination for the expanded fraction.
+ * @param end_of_buffer the end of `buffer`.
```
zslayton: I haven't written a lot of C recently, but I seem to recall `(char * buffer, int len)` being a more common way to represent a buffer than `(char * start, char * end)`. `ion_timestamp_to_string` uses this convention.

Either way, it's unclear from the doc comments whether 'the end of buffer' means it's the last writeable byte or if that's the `\0` terminator that can't be overwritten.

PR #157, opened 2020-01-01
(Tests) Qualify calls to isnan in tests

therapon:
*Description of changes:*

While updating my branch I got a compiler error for calls to `isnan` in test. Added qualifier `std::` to both calls. 


therapon:
> Approved, but the `ion-tests` commit should not change. Let's verify that isn't changed by the merge.

I will verify the changes to the git submodule for ion-tests. It did catch me by surprise when I created the PR and I looked at the diff for each file, it's empty. I think this is a side effect of my update from `upstream`, but I will make sure before pushing to `ion-c/master`

Issue #133, opened 2019-04-05
Consider alternatives to docopt


tgregg:
Docopt, which is used by the CLI for argument parsing, has strict toolchain requirements (see #131), has problems on Windows (see #95), and produces compiler warnings that point to possible memory leaks.

Some time should be spent to see if it's possible to quickly patch the docopt submodule to resolve these issues. Otherwise, consider replacing docopt with ionizer's argument parsing logic.

therapon:
I accidentally stumbled on this today [ArgTable](https://github.com/argtable/argtable3). Seems to be used for Unix, Windows and Mac.

PR #159, opened 2020-01-27
 (CLI) Replace docopt with argtable3 

therapon:
*Issue #, if available:*

Closes: #133 

*Description of changes:*
Moves CLI implementation to [argtable3](https://www.argtable.org/tutorial/) and removes doctop dependency. 

CLI output attempts to be 80 characters long. Only exception strings that are more than 80 characters long are commands and the CLI's options output. 

* Adds `help` as a subcommand that takes the name of another subcommand `process`, `extract`, `compare` and prints out the subcommands help information 
* output-type when passed an invalid string used to default to a value that is internal (not possible to be set via CLI). Added that as an option with the name `text` (vs `pretty`)

Some sample of the output
```
$ion help                                                                                                                                                                                     
Usage: 

	ion help [process|compare|extract]

Commands: 

extract    Extract the symbols from the given input(s) into a shared symbol 
           table with the given name and version.

compare    Compare all inputs (which may contain Ion streams and/or EventStreams) 
           against all other inputs using the Ion data model's definition of 
           equality. Write a ComparisonReport to the output.

process    Read the input file(s) (optionally specifying ReadInstructions or a 
           filter) and re-write in the format specified by --output.

help       Prints this general help. If provided a command, prints help specific
           to that command.

version    Prints version information about this tool.
```

```
$ion help process                                                                                     
Usage: 

	ion process [-o|--output=<file>] [-e|--error-report=<file>] [-f|--output-format=<type>] [-c|--catalog=<file>]... [-i|--imports=<file>]... [-p|--perf-report=<file>] [-F|--filter=<filter>] [-t|--traverse=<file>] <file> [<file>]...

Options: 

  -o, --output=<file>       Output file [default: stdout].
  -e, --error-report=<file> Error report file [default: stderr].
  -f, --output-format=<type> 
                            Output format, from the set (text | pretty | binary
                            | events | none). 'events' is only available with
                            the 'process' command, and outputs a serialized
                            EventStream representing the input Ion stream(s).
                            [default: pretty]
  -c, --catalog=<file>      Location(s) of files containing Ion streams of
                            shared symbol tables from which to populate a
                            catalog. This catalog will be used by all readers
                            and writers when encountering shared symbol table
                            import descriptors.
  -i, --imports=<file>      Location(s) of files containing list(s) of shared
                            symbol table import descriptors. These imports will
                            be used by writers during serialization. If a
                            catalog is available (see: --catalog), the writer
                            will attempt to match those import descriptors to
                            actual shared symbol tables using the catalog.
  -p, --perf-report=<file>  PerformanceReport location. If left unspecified, a
                            performance report is not generated.
  -F, --filter=<filter>     JQ-style filter to perform on the input stream(s)
                            before writing the result.
  -t, --traverse=<file>     Location of a file containing a stream of
                            ReadInstructions to use when reading the input
                            stream(s) instead of performing a full traversal.
  <file>                    Input file(s).

Examples: 

Read input.10n and pretty-print it to stdout:
	$ ion process input.10n

Read input.10n (using a catalog comprised of the shared symbol tables contained 
in catalog.10n) without re-writing, and write a performance report:
	$ ion process --output-format none --catalog catalog.10n --perf-report report input.10n

Read input.10n according to the ReadInstructions specified by instructions.ion 
and write the resulting Events to output.ion:
	$ ion process -o output.ion -f events -t instructions.ion input.10n

Read input1.ion and input2.10n and output to stdout any values in the streams 
that match the filter .foo:
	$ ion process --filter .foo input1.ion input2.10n

```


tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
-    return ion_cli_parse({ argv + 1, argv + argc });
-}
\ No newline at end of file
+
+
+
+
+
+
+
```
tgregg: Lots of extra whitespace.

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+    std::cout << std::endl << "Usage : " << std::endl;
+    PRINT_ION_COMMAND;
+    arg_print_syntaxv(stdout, args_table, "\n");
+    arg_print_glossary_gnu(stdout, args_table);
+}
+
+/**
+ * Check that each argtable instance was created and initialized.
+ */
+bool arg_tables_initalized() {
```
tgregg: Based on usage, it looks like this should be named something like `arg_tables_not_initialized`.

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+static const char *const OUTPUT_FILE_GLOSSARY = "Output file (default: stdout).";
+
+static const char *const ERR_SHORT_OPT = "e";
+static const char *const ERR_LONG_OPT = "error-report";
+static const char *const ERR_GLOSSARY = "Error report file (default: stderr).";
+
+static const char *const OUT_TYPE_SHORT_OPT = "f";
+static const char *const OUT_TYPE_LONG_OPT = "output-format";
+static const char *const TYPE_DATATYPE = "<type>";
+static const char *const OUT_TYPE_GLOSSARY1 = "Output format, from the set (text | pretty | binary | events | none).'events' is only available with the 'process' command, and outputs a serialized EventStream representing the input Ion stream(s).[default: pretty]";
```
pbcornell: add a space before `'events'` and `[default: pretty]`

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+// subcommand: compare
+static const char *const COMPARE_COMMAND_NAME = "compare";
+
+static const char *const COMPARE_TYPE_SHORT_OPT = "y";
+static const char *const COMPARE_TYPE_LONG_OPT = "comparison-type";
+static const char *const COMPARE_TYPE_GLOSSARY1 = "Comparison semantics to be used with the compare command, from the set (basic | equivs | non-equivs | equiv-timeline). Any embedded streams in the inputs are compared for EventStream equality. 'basic' performs a standard data-model comparison between the corresponding events (or embedded streams) in the inputs.'equivs' verifies that each value (or embedded stream) in a top-level sequence is equivalent to every other value (or embedded stream) in that sequence. 'non-equivs' does the same, but verifies that the values (or embedded streams) are not equivalent. 'equiv-timeline' is the same as 'equivs', except that when top-level sequences contain timestamp values, they are considered equivalent if they represent the same instant regardless of whether they are considered equivalent by the Ion data model.[default: basic]";
```
pbcornell: add a space before `'equivs'` and `[default: basic]`

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+//    subcommand: extract
+//    ion extract [--output <file>]
+//    [--error-report <file>]
+//    [--output-format <type>]
+//    (--symtab-name <name>)
+//    (--symtab-version <version>)
+//    ( - | <input_file>...)
+
+static struct arg_rex *xtract = arg_rex1(NULL, NULL, EXTRACT_COMMAND_NAME, NULL, REG_ICASE, NULL);
```
tgregg: Any reason not to just name all these "extract" vs "xtract"?

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+    cleanup:
     if (report.hasErrors()) {
         if (IERR_OK != ion_cli_write_error_report(&report, &common_args)) {
             std::cerr << "Error writing error report to " << common_args.error_report.contents << "." << std::endl;
         }
     }
+
+    exit:
+    // deallocate each non-null entry in each argtableoe
```
tgregg: argtableoe -> argtable

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+// Printing functions
+//
+
+void ion_print_process_examples() {
+    std::string s =
+            //--------------------------------------------------------------------------------
+            //<-                                80 chars                                    ->
+            //--------------------------------------------------------------------------------
+            "Read input.10n and pretty-print it to stdout.\n"
+            "\t $ ion process input.10n\n\n"
```
pbcornell: remove space before `$`

tools/cli/main.cpp
tgregg: Why are some of these suffixed with "GLOSSARY" while others have "GLOSSARY1"?

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+            "Read input.ion (using a catalog comprised of the shared symbol tables contained \n"
+            "in catalog.10n) without re-writing, and write a performance report.\n"
+            "\t$ ion process --output-format none --catalog catalog.10n --perf-report report -- input.10n\n\n"
```
pbcornell: is ` -- ` here intentional?  i can't tell what it means....

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+void ion_print_process_examples() {
+    std::string s =
+            //--------------------------------------------------------------------------------
+            //<-                                80 chars                                    ->
+            //--------------------------------------------------------------------------------
+            "Read input.10n and pretty-print it to stdout.\n"
+            "\t $ ion process input.10n\n\n"
+
+            "Read input.ion (using a catalog comprised of the shared symbol tables contained \n"
```
pbcornell: the command doesn't reference `input.ion`

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+
+            "compare    Compare all inputs (which may contain Ion streams and/or EventStreams) \n"
+            "           against all other inputs using the Ion data model's definition of \n"
+            "           equality. Write a ComparisonReport to the output.\n\n"
+
+            "process    Read the input file(s) (optionally, specifying ReadInstructions or a \n"
+            "           filter) and re-write in the format specified by --output.\n\n"
+
+            "help       Prints this general help. If provided a command, prints help specific\n"
+            "            to that command.\n\n"
```
tgregg: For consistency, could put the leading space on the previous line before \n so the columns align. (nitpick)

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
-    docopt::value val = it->second;
-    return val.isString() || val.isBool() || val.isLong() || val.isStringList(); // Otherwise, the value is Empty.
+void ion_print_compare_examples() {
+    std::string s =
+            //--------------------------------------------------------------------------------
+            //<-                                80 chars                                    ->
+            //--------------------------------------------------------------------------------
+            "Compare each stream in read_events.ion, input1.ion, and input2.10n against all \n"
+            "other streams in the set and output a ComparisonReport to \n"
+            "comparison_report.ion.\n"
```
pbcornell: looks like this might fit at the end of the previous line?

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+//
+// Printing functions
+//
+
+void ion_print_process_examples() {
+    std::string s =
+            //--------------------------------------------------------------------------------
+            //<-                                80 chars                                    ->
+            //--------------------------------------------------------------------------------
+            "Read input.10n and pretty-print it to stdout.\n"
```
pbcornell: for all of these examples, i'd prefer to see a `:` at the end of the description instead of a `.`

tools/cli/main.cpp
tgregg: (This comment applies to all examples) It looks like argtable3 requires a `=` between the option name and value, e.g. `--output-format=none` as opposed to `--output-format none`, is that correct?

If so, we'll need to update the examples and also [ion-test-driver](https://github.com/amzn/ion-test-driver).

tools/cli/main.cpp
therapon: I originally was using the argtable mechanism that does not wrap or format that required the strings to be split thus the suffix number. Since then I found a better that will auto-format and I missed the naming convention.

tools/cli/main.cpp
therapon: Fat fingers on my part.

tools/cli/main.cpp
therapon: Good catch. Thanks

tools/cli/main.cpp
therapon: I think that is a mistake. I took the text straight from the previous implementation. I will verify and fix.

therapon: It works both with as well as without 
```
therapon@u0164d560980059 > echo 'abc' | ./build/release/tools/cli/ion  process --output-format=binary -                                                    [~/Amazon/Work/GitHub/Ion/ion-c]
��遃և��abcq
therapon@u0164d560980059 > echo 'abc' | ./build/release/tools/cli/ion  process --output-format binary -                                                    [~/Amazon/Work/GitHub/Ion/ion-c]
��遃և��abcq

```

tools/cli/main.cpp
therapon: Yeap.

tools/cli/main.cpp
```diff
@@ -12,246 +12,859 @@
...
+}
+
+
+/**
+ * Given the number of file name and the file names as strings check if they contain the special
+ * string "-" that denotes STDIN. If so return false, esle true.
+ *
+ * @param count number of file names
+ * @param in_fnames file names as strings
+ * @return true if "-" is in `in_fnames`, false otherwise
```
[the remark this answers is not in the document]
therapon: Yeap it does .. sorry.

tools/cli/main.cpp
tgregg: Excellent!

therapon:
> Top-level items:
> 
>     * for a command with subcommands, `ion help` is producing more detail than i expect, and it repeats the following options 3 times:  `output`, `error-report`, `output-format`, and `catalog`.  the result makes this utility look more complex than it actually is.  i'd expect `ion help` to simply output the list of subcommands
>

For the behaviour of `ion help` I replicated what the CLI already was doing. However, before this PR the CLI did not deal with subcommands for `ion help`, i.e., `ion help process` was not implemented. I can go either way here. 


 
>     * some default are specified inside parens, others inside square brackets;  pick one [brackets are nice ;]

Changed. 

> 
>     * output for `ion help process` is grouped as `Options` and `Glossary`;  would prefer these groupings to be `Usage` and `Options` (as they are for `ion help`)
> 

Changed. 

> 
> Perhaps we're bumping up against some limits of the tool...?

No, it was me being influenced by the new tools terminology.

PR #177, opened 2020-06-30
Added static build targets.

zslayton:
*Description of changes:*

* Creates a static library build target for the CLI that does not have a `main()` function. This allows other programs to use the CLI's functionality while defining their own entry points.
* Makes `ion_events` a static library.
* Adds a static build target for `decNumber`.

### DIFF
--- a/decNumber/CMakeLists.txt
+++ b/decNumber/CMakeLists.txt
@@ -33,6 +33,21 @@ set_target_properties(decNumber
     PROPERTIES
         PUBLIC_HEADER "${DEC_PUB_HEADERS}")
 
+# ---
+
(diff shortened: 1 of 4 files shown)


zslayton:
Making `ion_events` a static library wasn't an issue on macOS, but it's causing the Linux builds to fail. I'll add a commit that creates an `ion_events_static` target alongside the `.so` instead of replacing it.

therapon:
> Should building the static targets be optional?

Working on that, got blocked due to windows builds failing on static lib. 
https://github.com/amzn/ion-c/compare/ts-unix-static

For now this gets the CLI work unblocked.

PR #180, opened 2020-07-31
Changes globals to use thread-local storage (TLS).

almann:
When working on Ion Rust bindings to Ion C, I found random crashes in the tests within that crate.  I realized that the tests were running multi-threaded and our use of mutable globals were (at least part) of the issue.  This change addresses that by introducing thread local storage modifiers to those particular globals.  The thread local macro is portable for C11 (which our builds don't yet target), GCC (and Clang), and MSVC.

* Adds the `THREAD_LOCAL_STORAGE` macro for portable declaration of TLS.
* Changes `g_ion_alloc_page_list` to be TLS.
  - This variable is mutated in `_ion_alloc_owner`.
* Changes the runtime initialized globals in `ion_int.h` to use TLS.
* Changes static globals for system symbol tables to use TLS.

I have built and run the tests on Ubuntu 18 GCC 7.5/Clang 10 and Windows MSVC 19.  I have also tested the associated Ion C bindings in rust on multi-threaded runs on Linux/Ubuntu for at least 10 minutes in a loop (previously it would fail pretty fast)--this is not a exhaustive test, but at least a reasonable litmus.


ionc/ion_symbol_table.c
```diff
@@ -211,8 +211,9 @@ iERR ion_symbol_table_get_system_table(hSYMTAB *p_hsystem_table, int32_t version
     iRETURN;
 }
 
-// HACK - TODO - hate this, needs to be fixed up for thread safety issues
-static ION_SYMBOL_TABLE *p_system_symbol_table_version_1 = NULL;
+// HACK - TODO - hate this
```
tgregg, line 214: I like how you left the "hate this" portion of the comment 👍.

PR #190, opened 2020-08-12
Fix issue causing lobs to skip some bytes when doing partial reads

### LINKED ISSUE #188: error calculating remaining bytes when reading from text lob stored in value_image
I ran into a problem trying to read an entire lob block after it had been loaded into the scanner value image from a previous get lob size call. I tracked down the issue to line 1925 in ion_reader_text.c where it updates the length of the remaining bytes.

The line currently reads: 
__text->_scanner._value_image.length -= remaining;_

but I think it should probably be setting the length to be equal to 'remaining' or just subtracting the bytes written.

It seems to work as I would expect with these changes in my local copy.

dlurton:
Fixes #188 

Also adds regression tests as well.

### DIFF
--- a/test/test_ion_text.cpp
+++ b/test/test_ion_text.cpp
@@ -17,6 +17,8 @@
 #include "ion_test_util.h"
 #include "ion_event_equivalence.h"
 
+#include "stdlib.h"
+
 TEST(IonTextSexp, ReaderHandlesNested)
 {
     const char* ion_text = "((first)(second))((third)(fourth))";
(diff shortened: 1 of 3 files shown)


test/test_ion_text.cpp
```diff
@@ -576,6 +578,103 @@ TEST(IonTextDecimal, FailsEarlyOnInvalidDecimal) {
...
+    ASSERT_EQ(EXPECTED_SIZE, total_bytes_read);
+    char* lob_text = (char*)bytes;
+    ASSERT_EQ(strcmp("This is a CLOB of text.", lob_text), 0);
+
+    free(bytes);
+    ION_ASSERT_OK(ion_reader_close(reader));
+}
+ 
+// regression test for https://github.com/amzn/ion-c/issues/188
+TEST(IonTextblob, CanFullyReadBlobUsingPartialReadsAfterReadingSize) {
```
tgregg, line 655: IonTextblob -> IonTextBlob

dlurton:
I've refactored a bit to DRY some and also added equivalent tests for Ion binary.  The same bug does not exist for the binary reader.

derivativerobmc:
thanks for fixing that so quickly

PR #212, opened 2020-10-30
Matches behavior between the text and binary readers after they're cr…

rgantt:
…eated, but before their position has been advanced.

This is just a parity gap between the text and binary ion readers, which in the normal case only impacts code with too-verbose diagnostics. In the worst case, it could be used by an SDK consumer to determine what kind of reader is currently being used, breaking encapsulation. Before the change, the unit test I added to the binary reader fails. After the change, it passes (without causing any other failures).

Now, the binary and text readers will both yield `type_none` in this situation.

https://issues.amazon.com/issues/ION-1226

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ionc/ion_reader_binary.c
+++ b/ionc/ion_reader_binary.c
@@ -486,7 +486,9 @@ iERR _ion_reader_binary_get_type(ION_READER *preader, ION_TYPE *p_value_type)
         // the next value. All other values must be positioned before the contents of the current value.
         if (binary->_value_type != tid_SYMBOL)
         {
-            FAILWITH(IERR_INVALID_STATE);
+            // This matches the behavior of the text reader
+            *p_value_type = tid_none;
+            SUCCEED();
         }
     }
 
--- a/test/test_ion_binary.cpp
+++ b/test/test_ion_binary.cpp
@@ -337,6 +337,16 @@ TEST(IonBinarySymbol, ReaderReadsIVMInsideAnnotationWrapper) {
     ION_ASSERT_OK(ion_reader_close(reader));
 }
 
+TEST(IonBinaryReader, UnpositionedReaderHasTypeNone) {
+    hREADER reader;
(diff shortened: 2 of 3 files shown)


tgregg:
Thanks! This basically looks good, but I'm wondering if we can just delete this entire block in `_ion_reader_binary_get_type`:

```
    if (!(preader->_eof) && binary->_state == S_BEFORE_TID)
    {
        // Non-null symbols are read up front, so a reader positioned on a symbol may be positioned before the TID of
        // the next value. All other values must be positioned before the contents of the current value.
        if (binary->_value_type != tid_SYMBOL)
        {
            // This matches the behavior of the text reader
            *p_value_type = tid_none;
            SUCCEED();
        }
    }
```

I'm wondering because I see that `binary->_value_type` is set to `tid_none` [when the value is reset](https://github.com/amzn/ion-c/blob/6f91da9624b7ec202505c879ce846cf3389c5964/ionc/ion_reader_binary.c#L85), which occurs between values and before the user calls `next`. And after the highlighted block, the function just returns `binary->_value_type`. Would you mind giving that a try? If it doesn't work we can go with the solution as currently proposed.

rgantt:
Good thinking. Easy enough to try!

When I remove the whole block, the new test I wrote fails a different way:

```
[----------] 1 test from IonBinaryReader
[ RUN      ] IonBinaryReader.UnpositionedReaderHasTypeNone
/Users/ryangant/workspaces/git/ion-c/test/test_ion_binary.cpp:347: Failure
      Expected: ((ION_TYPE)(-0x200))
      Which is: 0xfffffffffffffe00
To be equal to: type
      Which is: NULL
[  FAILED  ] IonBinaryReader.UnpositionedReaderHasTypeNone (0 ms)
[----------] 1 test from IonBinaryReader (0 ms total)
```

tgregg (approved):

Issue #224, opened 2021-02-04
`ion_int_from_long` never returns when MIN_INT64 is passed


zslayton:
In the following code, the call to `ion_int_from_long(&iint, MIN_INT64)` will never return:
```c
    // ...
    hWRITER writer;
    ion_writer_open_buffer(&writer, (BYTE *) buffer, (SIZE) 2048, &options);
    ION_INT iint;
    ion_int_init(&iint, &writer);
    ion_int_from_long(&iint, MIN_INT64);
```

When `ion_int_from_long` sets out to calculate the magnitude (i.e. the absolute value) of the provided value, [it checks to see if the value is negative and if so, negates it](https://github.com/amzn/ion-c/blob/1c4c6148a4b5d1099e46c62e9b7d4f13d87f0428/ionc/ion_int.c#L628-L630):
```c
    if ((is_neg = (value < 0)) == TRUE) {
        value = -value;
    }
```

This code fails when the provided `value` is `MIN_INT64`.

`MIN_INT64` is defined [here](https://github.com/amzn/ion-c/blob/d256918300dab1af831bd5c8db01a626688d1cd0/ionc/include/ionc/ion_types.h#L63-L64):
```c
#define MAX_INT64  0x7FFFFFFFFFFFFFFFLL
#define MIN_INT64  -0x7FFFFFFFFFFFFFFFLL-1
```

`MAX_INT64` is the largest value that can be stored in a signed 64-bit integer, `int64_t`. Notice that `MIN_INT64`'s magnitude is one greater than that of `MAX_INT64`. This means that the absolute value of `MIN_INT64` is too large to fit in an `int64_t`. Attempting to evaluate `-MIN_INT64` or `abs(MIN_INT64` is undefined behavior. In my build environment, the compiler treats it as a no-op.

In this corner case, `value = -value` does nothing and the following loop performs a series of right shifts on a value that (unexpectedly) still has its sign bit set to `1`:
```c
    ii_length = 0; 
    temp = value;
    while (temp) {
        temp >>= II_SHIFT; // <-- With the sign bit set, `temp` will never reach zero from right-shifting.
        ii_length++;
    }
```
Under normal circumstances, each right shift operation would drive `temp` closer to zero, ultimately ending the loop. However, with the sign bit set, each right shift populates the most significant bits in `temp` with `1`s, leading to an infinite loop.

zslayton:
Fixed by PR #225.

PR #225, opened 2021-02-04
Fixed issue #224: ion_int_from_long doesn't return

zslayton:
This PR addresses a corner case in `ion_int_from_long` that would cause an infinite loop when MIN_INT64 is passed in as the long value. A detailed description of the problem can be found in issue #224.

### DIFF
--- a/ionc/ion_int.c
+++ b/ionc/ion_int.c
@@ -40,6 +40,7 @@
 
 #include <decNumber/decNumber.h>
 #include "ion_internal.h"
+#include <stdlib.h>
 
 iERR ion_int_alloc(void *owner, ION_INT **piint)
 {
@@ -615,8 +616,9 @@ iERR ion_int_from_long(ION_INT *iint, int64_t value)
 {
     iENTER;
     SIZE  ii_length, digit_idx;
-    int64_t temp;
-    BOOL    is_neg;
+    // Stores the unsigned magnitude of the provided int64_t value. This variable must be
+    // unsigned to accommodate the absolute value of MIN_INT64, which requires 64 bits to store.
+    uint64_t magnitude;
 
     IONCHECK(_ion_int_validate_arg(iint));
     
@@ -625,25 +627,24 @@ iERR ion_int_from_long(ION_INT *iint, int64_t value)
         SUCCEED();
     }
 
-    if ((is_neg = (value < 0)) == TRUE) {
-        value = -value;
-    }
-
-    ii_length = 0; 
-    temp = value;
-    while (temp) {
-        temp >>= II_SHIFT;
+    ii_length = 0;
+    magnitude = (uint64_t) value;
+    while (magnitude) {
+        magnitude >>= II_SHIFT;
         ii_length++;
     }
 
+    // Reallocate iint's storage if it's not big enough to hold
+    // (ii_length * II_BITS_PER_II_DIGIT) bits.
     IONCHECK(_ion_int_extend_digits(iint, ii_length, TRUE));
 
-    for (digit_idx = iint->_len-1; value; digit_idx--) {
-        iint->_digits[digit_idx] = (II_DIGIT)(value & II_MASK);
-        value >>= II_SHIFT;
+    magnitude = (uint64_t) value;
+    for (digit_idx = iint->_len-1; magnitude; digit_idx--) {
+        iint->_digits[digit_idx] = (II_DIGIT)(magnitude & II_MASK);
+        magnitude >>= II_SHIFT;
     }
 
-    iint->_signum = is_neg ? -1 : 1;
+    iint->_signum = value < 0 ? -1 : 1;
 
     iRETURN;
 }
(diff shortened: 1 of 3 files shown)


tgregg:
The fix looks good, but can you please include the unit test?

zslayton:
> The fix looks good, but can you please include the unit test?

Polishing up my unit test has surfaced another problem. 😞 See #226. I'm going to try and roll a fix for that into this PR.

ionc/ion_int.c
```diff
@@ -1599,28 +1611,63 @@ SIZE _ion_int_abs_bytes_signed_length_helper(const ION_INT *iint)
...
     digits = iint->_digits;
     end    = digits + iint->_len;
-    while (digits < end) {
+
+    // iint's magnitude is stored in an array of 31-bit II_DIGIT values. Magnitudes that
+    // require 63 or 64 bits to represent will then take 3 II_DIGITS, only using 1 to 2
+    // bits from the most significant II_DIGIT.
+    const uint32_t max_digits_per_int64_t = 3;
+    const uint32_t whole_digits_per_int64_t = 2;
+    const uint32_t max_partial_digit_value = 3; // == 0b11, two populated bits
```
zslayton, line 1628: These constants went through a few iterations before I settled on this.

I tried re-calculating these values based on `sizeof(int64_t)` and `II_BITS_PER_II_DIGIT`, but it was pretty verbose:

```c
const uint32_t bits_per_int64_t = sizeof(int64_t) * II_BITS_PER_BYTE;
const float digits_per_int64_t = (float) bits_per_int64_t / (float) II_BITS_PER_II_DIGIT;
const uint32_t max_digits = ceilf(digits_per_int64_t);
const uint32_t whole_digits = floorf(digits_per_int64_t);
const uint32_t max_bits_in_partial_digit = bits_per_int64_t % II_BITS_PER_II_DIGIT;
const uint32_t max_partial_digit_value = (1<<max_bits_in_partial_digit) - 1;
```

While a compiler would've probably been able to reduce these to zero-cost constants at build time, the complexity it introduced for folks trying to read the code felt too expensive.

Given that we're using a type name that will always be 64-bits (`int64_t` instead of, say, `long`), this would only ever need to be reevaluated if the definition of `II_BITS_PER_II_DIGIT` (currently `31`) ever changed, which won't happen any time soon. The code above didn't seem to carry its own weight. I opted to just use the resulting values instead.

I also tried defining macros for these values in `ion_int.h`, but it felt like overkill; readers would have to track down their definitions in another file despite this being the only place they're ever used.

ionc/ion_int.c
```diff
@@ -625,25 +630,32 @@ iERR ion_int_from_long(ION_INT *iint, int64_t value)
...
     }
 
-    if ((is_neg = (value < 0)) == TRUE) {
-        value = -value;
+    is_negative = value < 0;
+    magnitude = (uint64_t) value;
+    if (is_negative) {
+        // This negates an unsigned value, which is well-defined behavior. Doing so handles the
+        // MIN_INT64 case: an int64_t whose absolute value is too large to be stored in an int64_t.
+        magnitude = -magnitude;
```
zslayton, line 638: Here's [a helpful StackOverflow answer](https://stackoverflow.com/questions/17313579/is-there-a-safe-way-to-get-the-unsigned-absolute-value-of-a-signed-integer-with#17313717) that includes the relevant text of the C spec.

ionc/ion_int.c
```diff
@@ -615,8 +616,12 @@ iERR ion_int_from_long(ION_INT *iint, int64_t value)
 {
     iENTER;
     SIZE  ii_length, digit_idx;
-    int64_t temp;
-    BOOL    is_neg;
+    // Stores the unsigned magnitude of the provided int64_t value. This variable must be
+    // unsigned to accommodate the absolute value of MIN_INT64, which requires 64 bits to store.
+    uint64_t magnitude;
+    // Used for shifting operations that consume the variable.
+    uint64_t temp_magnitude;
```
zslayton, line 623: `magnitude` is bit-shifted to zero twice. Given that it takes a small amount of expense to calculate it the first time, I introduced this temporary variable that could be reset to the value of `magnitude` between bit-shifting operations.

test/test_ion_integer.cpp
```diff
@@ -0,0 +1,58 @@
...
+    // Initialize the Ion integer, setting its owner to NULL.
+    IONCHECK(ion_int_init(&iint, NULL));
+    // Populate the Ion integer with the provided int64_t value.
+    IONCHECK(ion_int_from_long(&iint, value_in));
+    // Read the Ion integer's value back out into the output int64_t.
+    IONCHECK(ion_int_to_int64(&iint, value_out));
+    iRETURN;
+}
+
+TEST(IonInteger, IIntToInt64RoundTrip) {
```
tgregg, line 32: Should we also add a test that verifies overflow is detected when expected?

tgregg (approved):

tgregg (approved):

PR #227, opened 2021-02-09
Fixes a bug that caused an infinite loop when writing to a fixed-size buffer that is too small

### LINKED ISSUE #185: Small Buffer Causes Binary Writer Infinite Loop
The following test case causes an infinite loop--specifically for an over-small buffer that cannot possibly fit what is written.

```c
TEST(IonBinaryWriterBuffer, SmallBuffer) {
    iENTER;
    hWRITER writer = NULL;
    uint8_t buf[2];
    ION_WRITER_OPTIONS options = { 0 };
    options.output_as_binary = TRUE;
    ION_ASSERT_OK(ion_writer_open_buffer(&writer, buf, sizeof(buf), &options));
    ION_ASSERT_OK(ion_writer_write_int32(writer, 1));

    SIZE len;
    ION_ASSERT_FAIL(ion_writer_finish(writer, &len));
    ION_ASSERT_FAIL(ion_writer_close(writer));
}
```

tgregg:
*Issue #, if available:*

Fixes #185 

*Description of changes:*

* Fixes a bug that caused an infinite loop when writing to a fixed-size buffer that is too small.
* Adds a test to demonstrate how to use a user handler to grow the buffer when necessary.

### DIFF
--- a/test/test_ion_stream.cpp
+++ b/test/test_ion_stream.cpp
@@ -22,6 +22,7 @@
 #include <ionc/ion.h>
 #include "ion_helpers.h"
 #include "ion_test_util.h"
+#include "ion_assert.h"
 
 struct _test_in_memory_paged_stream_context {
     BYTE *data;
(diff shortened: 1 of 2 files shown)


test/test_ion_stream.cpp
```diff
@@ -103,3 +104,128 @@ TEST(IonStream, ContinuesOverPageBoundary) {
     ION_ASSERT_OK(ion_timestamp_equals(&expected, &ts, &is_equal, &g_IonEventDecimalContext));
     ASSERT_TRUE(is_equal);
 }
+
+TEST(IonStream, BufferTooSmall) {
+    iENTER;
+    hWRITER writer = NULL;
+    uint8_t buf[2]; // This buffer is to small to hold the output.
```
zslayton, line 111:
```suggestion
    uint8_t buf[2]; // This buffer is too small to hold the output.
```

zslayton (dismissed):

zslayton (approved):

PR #231, opened 2021-03-17
Update ion_platform_config.h

### LINKED ISSUE #230: Compiler Warning: extra tokens at end of #ifdef directive
The library uses the following construct:

```
 #ifdef __STDC_VERSION__ >= 201112L
```

which produces the mentioned warning. I think it should use the following instead:

```
 #if __STDC_VERSION__ >= 201112L
```

jobarr-amzn:
Fixes #230 by not misusing `#ifdef`

*Issue #, if available:* #230

*Description of changes:*

`#ifdef __STDC_VERSION__ >= 201112L` *might* work for some compilers but shouldn't by spec, whereas what C specifications I could find (C99, C0X, and the new C standard) all agree [in 6.10.1](http://c0x.coding-guidelines.com/6.10.1.html#1878) that:
>After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers are replaced with the pp-number `0`, and then each preprocessing token is converted into a token.


almann (approved):
I am almost certain that I am the source of this bug--so LGTM :ship:

jobarr-amzn:
So one of the Travis builds (GCC-4.9) is failing with errors like this: 

```
In file included from /home/travis/build/amzn/ion-c/ionc/include/ionc/ion.h:29:0,

                 from /home/travis/build/amzn/ion-c/ionc/ion_internal.h:33,

                 from /home/travis/build/amzn/ion-c/ionc/decQuadHelpers.c:17:

/home/travis/build/amzn/ion-c/ionc/include/ionc/ion_int.h:131:1: error: ‘__thread’ before ‘extern’

 THREAD_LOCAL_STORAGE ION_INT_GLOBAL BOOL            g_ion_int_globals_initialized; // NOTE: this is initialized to 0 according to C standard.
```

I believe that the fix #228 is correct based on:
https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Thread-Local.html#Thread-Local
https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html

README.md
```diff
@@ -44,5 +44,11 @@ currently specified in the `.gitmodules` file.
 For detailed walkthroughs of git submodule usage, see the
 [Git Tools documentation](https://git-scm.com/book/en/v2/Git-Tools-Submodules).
 
+## Building the Library
+Use the provided scripts `build-release.sh` and `build-debug.sh`. Ensure that `cmake` is installed first.
```
zslayton, line 48: I suggest rephrasing this to:
* Put the "first" step at the beginning 
* Clarify how to choose which build script a user should run

```suggestion
Ensure that `cmake` is installed. To create an optimized build without debug symbols, run the `build-release.sh` script. To create an unoptimized build that includes debug symbols, use `build-debug.sh`.
```

Is there a minimum version of `cmake`? Should we have a `dependencies` section that spells out platform and compiler versions? (Not necessarily authored in this PR, we can open an issue to track it.)

jobarr-amzn:
This removes about 2000 lines of compiler warnings, compare these two builds starting from ~ line 475
old: https://travis-ci.com/github/amzn/ion-c/jobs/491885577
new: https://travis-ci.com/github/amzn/ion-c/jobs/492126856

README.md
```diff
@@ -44,5 +44,11 @@ currently specified in the `.gitmodules` file.
...
+
+### On macOS
+`cmake` can be installed using [Homebrew](https://brew.sh/): `brew install cmake`
```
zslayton, line 51: This is good information but makes me wonder if we need to provide corresponding instructions for Linux/Windows?

zslayton (approved):

README.md
jobarr-amzn: I'm taking a stone soup approach here :) and general install instructions for linux are "install from your package manager whatever it is and how do you not have cmake on linux?" right?

README.md
jobarr-amzn: I don't know.

README.md
zslayton: > general install instructions for linux are "install from your package manager whatever it is and how do you not have cmake on linux?" right?

Linux has newbies too! And not every linux distro comes with `build-essentials` or equivalent preinstalled. It never hurts to have friendly/complete documentation.

README.md
zslayton: I'd be happy with a "known to work on" section.

Issue #139, opened 2019-10-01
Support 32-bit floats.


tgregg:
Currently, the binary reader [raises IERR_INVALID_BINARY](https://github.com/amzn/ion-c/blob/master/ionc/ion_binary.c#L389) for all float values that don't declare length 8. However, 32-bit floats (length 4) are also allowed.

Read support should be prioritized. Write support can also be added.

tgregg:
Read support has been added via `ion_reader_read_double` in #155 

Leaving this issue open for the remaining task: adding a writer API to write 32-bit floats.

abhinuvpitale:
Can I try the writing the writer API?

tgregg:
@abhinuvpitale Of course!

nikhilsr:
Is there a timeline for support of writing float values yet?

jobarr-amzn:
@nikhilsr For clarification, `ion-c` presently supports writing C float types as [Ion floats](https://amzn.github.io/ion-docs/docs/float.html), but they are written using the IEEE-754 `binary64` encoding. 

This issue is about adding a writer API to support writing floating point types using the fewer bits of the IEEE-754 `binary32` encoding.

We presently have no timeline for the latter, but I'll take a look.

jobarr-amzn:
This should be resolved by PR #232 . A future change will add an option to enable writing `double` values as 32-bit floats when it's losslessly possible.

PR #232, opened 2021-04-14
Add support for writing 32-bit floats

jobarr-amzn:
*Issue #, if available:* #139

*Description of changes:* Adds support for writing 32-bit floats via a new method exposed in `ion_writer.h`:

```c
ION_API_EXPORT iERR ion_writer_write_float          (hWRITER hwriter, float value);
```

Also adds unit tests for the same values used to verify 32-bit float read support, and amends the `assertBytesEqual` helper method to yield more useful feedback during failures. 


ionc/ion_binary.c
```diff
@@ -91,7 +91,19 @@ int ion_binary_len_var_int_64( int64_t value )
...
+{
+    int len = 0;
+    if (value == 0 && !ion_float_is_negative_zero(value)) {
+        len = 0;
+    }
+    else {
+        len = sizeof(value); // floats are IEEE 754 32 bit binary floating point
+    }
+    return len;
+}
```
almann, line 104: It is regrettable what we have a bunch of copy-paste here--unfortunately, the only real alternative is a macro, but I am not sure it is worth it

ionc/ion_binary.c
```diff
@@ -535,6 +547,39 @@ iERR ion_binary_write_float_value( ION_STREAM  *pstream, double value )
...
+    }
+
+    // now write the bytes out from most to least significant
+    // to the output stream
+    ASSERT((pb + 1) == image);
+
+    IONCHECK(ion_binary_write_byte_array(pstream, image, 0, UINT_32_IMAGE_LENGTH));
+
+    iRETURN;
+}
```
almann, line 581: The copy-paste here might be worth the macro...

ionc/ion_writer_impl.h
```diff
@@ -330,6 +331,7 @@ iERR _ion_writer_binary_write_typed_null(ION_WRITER *pwriter, ION_TYPE type);
 iERR _ion_writer_binary_write_bool(ION_WRITER *pwriter, BOOL value);
 iERR _ion_writer_binary_write_int64(ION_WRITER *pwriter, int64_t value);
 iERR _ion_writer_binary_write_ion_int(ION_WRITER *pwriter, ION_INT *iint);
+iERR _ion_writer_binary_write_float(ION_WRITER *pwriter, float value);
 iERR _ion_writer_binary_write_double(ION_WRITER *pwriter, double value);
```
almann, line 335: There is a weird asymmetry between these APIs and the `_float32` and `_float64` names.  I don't feel super strong about it, but I am curious as to your thoughts about this.

test/test_ion_binary.cpp
```diff
@@ -440,6 +440,10 @@ TEST(IonBinaryFloat, ReaderSupports32BitFloats) {
     int pos_inf_bits = 0x7F800000;
     float pos_inf = *((float *)&pos_inf_bits);
 
+    // See https://amzn.github.io/ion-docs/docs/binary.html#4-float
+    // "If L is 0, then the the value is 0e0 and representation is empty."
+    test_ion_binary_reader_supports_32_bit_floats((BYTE *) "\xE0\x01\x00\xEA\x40", 5, 0.);
```
almann, line 445: I think the zero case is good, but has nothing to do with 32-bit-ness per se...

test/test_ion_binary.cpp
```diff
@@ -547,3 +551,36 @@ TEST(IonBinaryBlob, CanFullyReadBlobUsingPartialReads) {
...
+
+    // ion-c prefers to write positive zero as a zero-length float
+    // see: https://amzn.github.io/ion-docs/docs/binary.html#4-float
+    test_ion_binary_writer_supports_32_bit_floats(  0., "\xE0\x01\x00\xEA\x40", 5);
+    test_ion_binary_writer_supports_32_bit_floats( -0., "\xE0\x01\x00\xEA\x44\x80\x00\x00\x00", 9);
+    test_ion_binary_writer_supports_32_bit_floats( 4.2, "\xE0\x01\x00\xEA\x44\x40\x86\x66\x66", 9);
+    test_ion_binary_writer_supports_32_bit_floats(-4.2, "\xE0\x01\x00\xEA\x44\xC0\x86\x66\x66", 9);
+
+    test_ion_binary_writer_supports_32_bit_floats(neg_inf, "\xE0\x01\x00\xEA\x44\xFF\x80\x00\x00", 9);
+    test_ion_binary_writer_supports_32_bit_floats(pos_inf, "\xE0\x01\x00\xEA\x44\x7F\x80\x00\x00", 9);
```
almann, line 581: Maybe add a case for `NaN`?

almann (dismissed):
Overall, the copy-paste makes me sad, but I am not sure macros are the right solution (...now you have two problems...).  Minor comments below around tests and naming.

ionc/ion_binary.c
jobarr-amzn: Agreed :(

test/test_ion_binary.cpp
jobarr-amzn: I'll look into this...

ionc/ion_binary.c
jobarr-amzn: I was going to say that I disagreed because of the number of parameters it would have to have:
* type of `intvalue`
* size of image
* expected size of image

Actually that's not so many, after all. Huh...

Issue #181, opened 2020-07-31
Add Sentinel for Pool Allocator Not Initialized


almann:
As part of fixing #180, there was a line of code that appears to be a bug:

https://github.com/amzn/ion-c/blob/a9af6be3f2ed38d4775257adfa5672c845cac361/ionc/ion_allocation.c#L39-L54
```
void *_ion_alloc_owner(SIZE len)
{
    void                 *owner;
    ION_ALLOCATION_CHAIN *new_chain;

    if (g_ion_alloc_page_list.page_size == ION_ALLOC_PAGE_POOL_PAGE_SIZE_NONE) {
        ion_initialize_page_pool(ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE, ION_ALLOC_PAGE_POOL_DEFAULT_LIMIT);
    }

    new_chain = _ion_alloc_block(len);
    if (!new_chain) return NULL;

    owner = _ion_alloc_with_owner_helper(new_chain, len, FALSE);

    return owner;
}
```

Specifically, `g_ion_alloc_page_list.page_size == ION_ALLOC_PAGE_POOL_PAGE_SIZE_NONE` should not be used as a sentinel for *page pool requires initialization*, because it is currently used to mean *don't use page pool*.  As a result the library today always uses the page pool.

We should create the right sentinel for this (probably make it per-thread configurable, since the pool is now thread-local), and also find other usages of `ion_initialize_page_pool` that might be doing the same.

PR #236, opened 2021-04-26
Fix g_ion_alloc_page_list initialization issue.

barczpe:
g_ion_alloc_page_list sometimes not initialized:
    - when symbol table is created first.
    - in multi-threaded use case, when ionc stream is initialized on one thread, and used from another one.
The initialization function is called with constant values, so the initialization list of g_ion_alloc_page_list is changed for using these constants.
Related to https://github.com/amzn/ion-c/issues/181

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ionc/ion_allocation.c
+++ b/ionc/ion_allocation.c
@@ -41,10 +41,6 @@ void *_ion_alloc_owner(SIZE len)
     void                 *owner;
     ION_ALLOCATION_CHAIN *new_chain;
 
-    if (g_ion_alloc_page_list.page_size == ION_ALLOC_PAGE_POOL_PAGE_SIZE_NONE) {
-        ion_initialize_page_pool(ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE, ION_ALLOC_PAGE_POOL_DEFAULT_LIMIT);
-    }
-
     new_chain = _ion_alloc_block(len);
     if (!new_chain) return NULL;
 
@@ -224,7 +220,8 @@ void _ion_free_block(ION_ALLOCATION_CHAIN *pblock)
 void ion_initialize_page_pool(SIZE page_size, int free_page_limit)
 {
     // once the page list is in use, you can't change your mind
-    if (g_ion_alloc_page_list.page_size != ION_ALLOC_PAGE_POOL_PAGE_SIZE_NONE)
+    // This is changed, because g_ion_alloc_page_list.page_size is initialized to ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE.
+    if (g_ion_alloc_page_list.page_size == ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE)
     {
         return;
     }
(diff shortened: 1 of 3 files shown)


tgregg:
Thanks for the contribution. I have a few questions.

1. Won't this change make it so that it is impossible to configure the page size via `ion_initialize_page_pool`?
2. Doesn't it also make it impossible to disable the page pool (which appears to be an existing defect, but one that this change perpetuates)?

barczpe:
Hi Greg,

I also changed the ion_initialize_page_pool function. It can be used to
configure the page size as long as it contains the default value.
Just like before. The only difference is that the default value is a useful
value, which can be used without the initialization.

I am not sure what do you mean by disabling the page pool. The default -1
does not really disabled it.
It seems to me, that the pool is used even if it is not initialized, or
used on a wrong way.

The reason I made this change is that in my application the memset()
function call overwrote the memory,
( if (ptr) memset(ptr, 0, length); line at the and of
_ion_alloc_with_owner_helper() )
because for some reason the page pool was not initialized.
These changes solved this issue.

However, I do not have a deep knowledge of the source, so this may not be
the proper solution.

Best regards,
Peter

ionc/ion_allocation.c
```diff
@@ -224,7 +220,8 @@ void _ion_free_block(ION_ALLOCATION_CHAIN *pblock)
 void ion_initialize_page_pool(SIZE page_size, int free_page_limit)
 {
     // once the page list is in use, you can't change your mind
-    if (g_ion_alloc_page_list.page_size != ION_ALLOC_PAGE_POOL_PAGE_SIZE_NONE)
+    // This is changed, because g_ion_alloc_page_list.page_size is initialized to ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE.
+    if (g_ion_alloc_page_list.page_size == ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE)
```
tgregg, line 224: Won't this always be true (note: this is `==`, not `!=` as before)? `g_ion_alloc_page_list.page_size` is statically initialized to `ION_ALLOC_PAGE_POOL_DEFAULT_PAGE_SIZE` as part of this change. And I don't think the solution is just to make this `!=` because that doesn't allow us to distinguish between the default and a user-selected value that happens to be the same as the default.

Now, this probably isn't a big deal since `ion_alloc.h` isn't actually part of the public includes, so users can't really call `ion_initialize_page_pool` manually anyway. It seems like it would be desirable for this to be configurable, either directly by calling `ion_initialize_page_pool` or indirectly via `ION_WRITER_OPTIONS`. However, we don't need to solve that as part of this change.

I would at least like to have a comment in here that acknowledges that `ion_initialize_page_pool` does nothing so that I can create an issue to look into addressing that, though.

barczpe:
Yes, you are right, it should be !=. (I do not remember why it is changed,
sorry.)

Can you fix this, while you are merging the change, or shall I fix it in
the pull request?

However, the problem is, that instead of checking the page_size, we should
check if ion_initialize_page_pool() is called before allocating pages.

Thanks,
Peter

tgregg (approved):
I'm actually going to merge this as-is (with `ion_initialize_page_pool()` doing nothing). Then I will create an issue for determining how to best support allowing users to configure the page pool and add a TODO comment in the code.

PR #244, opened 2021-05-21
adds changes for correctly setting stream position when error occurs while reading binary int64

desaikd:
*Description of changes:*
This PR works on setting stream position correctly when error occurs in reading binary int64 value.

*Changes:*
- Changed reader method for binary int64 value to set the position of stream back to the start of value in order to not consume it in error cases.

*Test:*
- Added tests for checking the edge case(integer of size 8 bytes) that doesn't fit in int64 throwing overflow error and then reading it as big integer.

### DIFF
--- a/ionc/ion_reader_binary.c
+++ b/ionc/ion_reader_binary.c
@@ -969,7 +969,7 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
 {
     iENTER;
     ION_BINARY_READER *binary;
-    int                tid, len;
+    int                tid, len, i;
     uint64_t           unsignedInt64 = 0;
     BOOL               is_negative;
     BOOL               is_null;
@@ -1006,7 +1006,17 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
     IONCHECK(ion_binary_read_uint_64(preader->istream, len, &unsignedInt64));
 
     is_negative = (tid == TID_NEG_INT)? TRUE: FALSE;
-    IONCHECK(cast_to_int64(unsignedInt64, is_negative, p_value));
+    iERR result = cast_to_int64(unsignedInt64, is_negative, p_value);
+
+    // if an error is thrown reset reader stream's current pointer to the beginning of current value again
+    if(result) {
+        i = len;
+        while(i--) {
+            preader->istream->_curr--;
+        }
+        FAILWITH(result);
+    }
+
     if (is_negative && *p_value == 0) {
         FAILWITH(IERR_INVALID_BINARY);
     }
--- a/ionc/ion_reader_binary.c
+++ b/ionc/ion_reader_binary.c
@@ -969,7 +969,7 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
 {
     iENTER;
     ION_BINARY_READER *binary;
-    int                tid, len, i;
+    int                tid, len;
     uint64_t           unsignedInt64 = 0;
     BOOL               is_negative;
     BOOL               is_null;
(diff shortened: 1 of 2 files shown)


ionc/ion_reader_binary.c
```diff
@@ -1006,7 +1006,17 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
     IONCHECK(ion_binary_read_uint_64(preader->istream, len, &unsignedInt64));
 
     is_negative = (tid == TID_NEG_INT)? TRUE: FALSE;
-    IONCHECK(cast_to_int64(unsignedInt64, is_negative, p_value));
+    iERR result = cast_to_int64(unsignedInt64, is_negative, p_value);
+
+    // if an error is thrown reset reader stream's current pointer to the beginning of current value again
+    if(result) {
```
almann, line 1012: I would write this as `result != iERR_OK`

ionc/ion_reader_binary.c
```diff
@@ -1006,7 +1006,17 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
...
+        i = len;
+        while(i--) {
+            preader->istream->_curr--;
```
almann, line 1015: I don't think you need a loop, can't you just `preader->istream->_curr -= len;`?

test/test_ion_binary.cpp
```diff
@@ -402,6 +402,38 @@ TEST(IonBinaryInt, ReaderRejectsNegativeZeroMixedIntTwoByte) {
...
+    ion_int_to_char(&big_int_expected, (BYTE *)int_str, str_len, &written);
+
+    // compare string representation of the value
+    ASSERT_STREQ(actual_value, int_str);
+}
+
+TEST(IonBinaryReader, ReaderThresholdForInt64) {
+    test_ion_binary_reader_threshold_for_int64((BYTE *)"\xE0\x01\x00\xEA\x28\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 13, "18446744073709551615");
+    test_ion_binary_reader_threshold_for_int64((BYTE *)"\xE0\x01\x00\xEA\x39\x00\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 14, "-18446744073709551615");
+}
```
almann, line 435: Make these two different cases.  Also why is the second case a size `9` negative integer?  I don't think that triggers the case you are testing.

almann (changes requested):

ionc/ion_reader_binary.c
```diff
@@ -1006,7 +1006,17 @@ iERR _ion_reader_binary_read_int64(ION_READER *preader, int64_t *p_value)
     IONCHECK(ion_binary_read_uint_64(preader->istream, len, &unsignedInt64));
 
     is_negative = (tid == TID_NEG_INT)? TRUE: FALSE;
-    IONCHECK(cast_to_int64(unsignedInt64, is_negative, p_value));
+
+    {
+        iERR cast_result = cast_to_int64(unsignedInt64, is_negative, p_value);
+
+        // if an error is thrown reset reader stream's current pointer to the beginning of current value again
+        if (cast_result :!= IERR_OK) {
```
almann: What is the `:!=`?  Is that a typo?

test/test_ion_binary.cpp
```diff
@@ -402,6 +402,41 @@ TEST(IonBinaryInt, ReaderRejectsNegativeZeroMixedIntTwoByte) {
...
+    ASSERT_STREQ(actual_value, int_str);
+}
+
+TEST(IonBinaryReader, ReaderPositiveThresholdForInt64) {
+    test_ion_binary_reader_threshold_for_int64((BYTE *)"\xE0\x01\x00\xEA\x28\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 13, "18446744073709551615");
+}
+
+TEST(IonBinaryReader, ReaderNegativeThresholdForInt64) {
+    test_ion_binary_reader_threshold_for_int64((BYTE *)"\xE0\x01\x00\xEA\x38\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 13, "-18446744073709551615");
+}
```
almann, line 438: Very minor, but I think variants of the test with `2 ** 63` is interesting `'\x80\x00\x00\x00\x00\x00\x00\x00'` should fail for positive int, but not negative int (since `-2 ** 63` fits as two's complement representation `0x8000000000000000`).

almann (approved):
:shipit:

almann (dismissed):
See previous comment around testing, but otherwise this looks good.

almann (approved):
:ship: it!

PR #247, opened 2021-06-04
Bypass event processing in the cli tool if both the input and output are value streams.

gergodocziaim:
*Description of changes:*

At our company we are often dealing with binary Amazon ion files that are a few gigabytes large.
Sometimes we want to convert them to text for manual inspection, and it would be convenient to use the provided CLI tool for this purpose. Basically we would like to do this: `ion process -f pretty -o output.txt large_binary_file.ion`

Unfortunately the tool currently tries to load the whole input file into memory and even multiplies its size by converting it into an event stream. For larger files this requires more memory than is available in a typical desktop computer.

I've noticed this comment in cli.cpp:
```
// Would be nice to use this (especially for performance testing), but having to peek at the first value in the
// stream (to identify $ion_event_stream) rules it out.
//IONCHECK(ion_writer_write_all_values(writer_context->writer, reader_context.reader));
```

I think I could make it work by converting only the first value into events in ion_event_stream_is_event_stream, writing it back calling ion_event_stream_write_all, and processing the rest with ion_writer_write_all_values as the comment suggests.
The only unit test that failed is IonCli.ErrorIsConveyed. It looks like it parses "false" from "false::23" and reports error only when it reaches the first colon at index 5. The original implementation reported error immediately at index 0. I modified the test for now, but I'm not sure which one is considered the correct behavior.

Note that this does not solve the memory usage problem when the input or the output is an event stream. Solving that would require a much deeper refactor, and since that use case is not really important for us, I did not attempt to do it.

### DIFF
--- a/test/test_ion_cli.cpp
+++ b/test/test_ion_cli.cpp
@@ -242,7 +242,7 @@ TEST(IonCli, ErrorIsConveyed) {
     test_ion_cli_process(test_file.c_str(), IO_TYPE_FILE, &command_output, &report);
     ASSERT_TRUE(report.hasErrors());
     ASSERT_FALSE(report.hasComparisonFailures());
-    ASSERT_EQ(0, command_output.length);
+    ASSERT_EQ(5, command_output.length);
     test_ion_cli_assert_error_equals(&report.getErrors()->at(0), ERROR_TYPE_READ, IERR_INVALID_SYNTAX, test_file);
     ion_cli_free_command_output(&command_output);
 }
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -277,19 +277,8 @@ iERR ion_cli_write_input(IonCliCommonArgs *args, IonEventWriterContext *writer_c
     IonCliReaderContext reader_context;
     IonEventStream stream(input->contents);
     IONREPORT(ion_cli_open_reader(input, catalog, &reader_context, &stream, result));
-    if (args->output_format == OUTPUT_TYPE_EVENTS) {
-        // NOTE: don't short-circuit on read failure. Write as many events as possible.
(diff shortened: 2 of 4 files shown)


cheqianh (commented):
Thanks for the contribution!

[ion-test-driver](https://github.com/amzn/ion-c/runs/2748292125?check_suite_focus=true) failed below due to ion-c read/write performance changed.
&nbsp;
Current change had a different output of reading/writing file [intWithLeadingZeros.ion](https://github.com/amzn/ion-tests/blob/master/iontestdata/bad/intWithLeadingZeros.ion) with text `[ 007 ]` by command `ion process -f events -o output intWithLeadingZeros.ion`.

**Previous output:** 
```.c
// It stopped after container_start '[' due to ion spec only allows one leading 0.
$ion_event_stream
{
  event_type: CONTAINER_START,
  ion_type: LIST,
  depth: 0
}
```
**Output after change:**
```.c
$ion_event_stream
```


As mentioned [here](https://github.com/amzn/ion-c/pull/247/files#diff-463707bc0f00fe891436d79ece50ab59da1082e5a2c84f0d9d073bbd8f7bc47bL281). we tried to write as many events as possible while read fails. Could you take a look at this?

gergodocziaim:
> As mentioned [here](https://github.com/amzn/ion-c/pull/247/files#diff-463707bc0f00fe891436d79ece50ab59da1082e5a2c84f0d9d073bbd8f7bc47bL281). we tried to write as many events as possible while read fails. Could you take a look at this?

Thanks, I'll look into it.

gergodocziaim:
I've fixed it, it passes the checks now.

tgregg (approved):
@gergodocziaim This is a really nice improvement. Thanks for the contribution!

PR #249, opened 2021-06-11
Bugfix negative zero for decimal to int

marcbowes:
*Issue #, if available:*

Related to https://github.com/amzn/ion-rust/issues/83.

*Description of changes:*

Before this commit, -0d0 to int would lose the sign. This is because the
special case for zero triggers (negative zero **is** zero) but writes
out the wrong integer (`_ion_int_zero` has a signum=0).

A note on testing: I couldn't find an obvious place to test this, but I did test it manually as part of my changes to ion-rust. Importantly, after this change I get back an IonInt that has a signum of -1. It looks like this:

```
IonIntPtr { value: 0x7fe6b9600330 }, sign -1
```

### DIFF
--- a/ionc/ion_int.c
+++ b/ionc/ion_int.c
@@ -680,13 +680,14 @@ iERR ion_int_from_decimal(ION_INT *iint, const decQuad *p_value, decContext *con
         FAILWITH(IERR_INVALID_ARG);
     }
 
-    // special case since zero is so common (and quick to test and set)
-    if (decQuadIsZero(p_value)) {
+    is_neg = decQuadIsSigned(p_value);
+
+    // special case since positive zero is so common (and quick to test and set)
+    if (decQuadIsZero(p_value) && !is_neg) {
         IONCHECK(_ion_int_zero(iint));
         SUCCEED();
     }
 
-    is_neg = decQuadIsSigned(p_value);
     decQuadCopyAbs(&temp1, p_value);
 
     decimal_digits = decQuadDigits(&temp1);


tgregg:
Why does `-0` need to be represented using an `ION_INT`? It's not clear to me that `ION_INT` should support that. Looking for precedent in a different "big integer" implementation, I found that Java's BigInteger treats `-0` as `0`.

```java
BigInteger integer = new BigInteger("-0");
System.out.println(integer); // prints 0
System.out.println(integer.signum()); // prints 0
```

Also note that binary Ion refuses to represent a `-0` integer.

marcbowes:
You're definitely right that many other languages don't seem to support this. The use-case is that I'm trying to chase down a -0d0 bug in ion-rust and it currently uses the integer conversion to extract a decimal coefficient. 

https://github.com/amzn/ion-rust/blob/main/ion-c-sys/src/lib.rs#L421

tgregg:
Can you use [`ion_decimal_is_negative`](https://github.com/amzn/ion-c/blob/master/ionc/include/ionc/ion_decimal.h#L242) in ion-c-sys if the coefficient is 0?

marcbowes:
`-0d0` is loaded using a quad, so

```
ION_DECIMAL_UTILITY_API_BUILDER(ion_decimal_is_negative, decQuadIsNegative, decNumberIsNegative);
```

lands up calling `decQuadIsNegative`. Which is *different* to `decQuadIsSigned` (the API this PR calls). `isNegative` is only true if the value is less than zero (which it isn't).

marcbowes:
For ref. here is the same docs for the quad case (I know it says double, it's the same though):
http://speleotrove.com/decimal/dnfloat.html:
> decDoubleIsNegative(x)
>Returns an unsigned integer (uint32_t) which will be 1 if x is less than zero and not a NaN, or 0 otherwise.

tgregg:
I think we should add `ion_decimal_is_signed`, which would call `decQuadIsSigned` for decQuad and `decNumberIsNegative` for decNumber. If my understanding is correct, the result of calling that function would not depend on the underlying representation. Then we should document the difference between the functions with respect to zero. I'd then be okay with deprecating `ion_decimal_is_negative`.

tgregg:
Closing this in favor of #252

PR #251, opened 2021-06-16
Fixes an issue where offset capture of a multi-field struct was incor…

rgantt:
Fixes an issue where offset capture of a multi-field struct was incorrectly returning field name positions instead of value positions

*Description of changes:*

This adds a condition to the resetting of the `value_start` position during the `_next` operation to bring the behavior of the text reader in line with the binary reader when doing offset capture.

### DIFF
--- a/test/test_ion_reader_seek.cpp
+++ b/test/test_ion_reader_seek.cpp
@@ -721,16 +721,10 @@ TEST_P(TextAndBinary, ReaderPopulatesStructFieldsOnSeek) {
 
     ION_ASSERT_OK(ion_reader_next(reader, &type));
     ASSERT_EQ(tid_SYMBOL, type);
-
-    // Real code would probably capture these, but it doesn't affect the issue
-    //ION_ASSERT_OK(ion_reader_get_field_name(reader, &read_field1));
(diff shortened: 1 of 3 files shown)


test/test_ion_reader_seek.cpp
```diff
@@ -679,4 +679,81 @@ TEST_P(TextAndBinary, ReaderHandlesInitialUnannotatedContainerValueOffsetSeek) {
...
+    ION_ASSERT_OK(ion_reader_read_string(reader, &read_val2));
+    char *cread_val2 = ion_string_strdup(&read_val2);
+
+    ION_ASSERT_OK(ion_reader_close(reader));
+
+    // Assert:
+
+    // Easy assertions: there's only one value, "value," and we should have read it both times
+    assertStringsEqual((char *)value.value, cread_val1, strlen(cread_val1));
+    assertStringsEqual((char *)value.value, cread_val2, strlen(cread_val2));
```
tgregg, line 758: This is probably good enough, but I would feel better about it if the values weren't the same. That way we know this won't happen to pass if the second seek somehow leaves the reader positioned at the first value.

tgregg (commented):
Looks good, minor comment below.

tgregg (approved):

PR #254, opened 2021-10-08
Migrates CI build from Travis to Github Actions

popematt:
**Issue #, if available:**

N/A

**Description of changes:**

Migrates CI build from Travis to Github Actions

In doing so, we change the compiler versions and build platforms a bit (details below). In order to get it to build on AL2012, I had to lower the cmake minimum version from 3.10 to 3.6—everything seems to still build just fine though.

You can see a successful run [here](https://github.com/popematt/ion-c/actions/runs/1319080048).

Here are the combinations of OS and compiler that are being checked in the new GH Actions workflow.

| OS | Compiler |
| --- | --- |
| Amazon Linux | clang-3.6.2 |
| Amazon Linux | gcc-4.8.5 |
| Amazon Linux 2 | clang-11.0.1 |
| Amazon Linux 2 | gcc-7.3.1 |
| Ubuntu 20.04 * | clang-11.0.0 |
| Ubuntu 20.04 * | gcc-9.3.0 |
| MacOS 10.15 * | AppleClang-12.0.0 |
| MacOS 10.15 * | gcc-11.2.0 ^ |

_*These are not fixed to specific versions. They use "latest" and will be updated sometime following a new LTS release of that OS._
_^This is the only compiler that is actually fixed to a specific version, though in practice, the Amazon Linux compiler versions are unlikely to change._


popematt:
> Very nice. While we're here: does Github Actions support Windows now? We still have an AppVeyor run for Windows, but it we can move that to Github Actions too, that would be great.

Yes, GH Actions has support for [3 versions of Windows Server](https://github.com/actions/virtual-environments#user-content-available-environments). I'll create an issue for us to try to migrate from AppVeyor to GH Actions.

PR #258, opened 2021-11-23
fix #221 - Ion binary floating point exception

### LINKED ISSUE #221: Ion binary crashes with floating point exception
Running build-release.sh to install creates ion and ionizer.
Going into ion-c/build/release/tools/cli

Running 
```
ion process [filename]
```

cause floating point exception for some inputs such as 
```
[[1, -12D516753754312305], [-629107941, 255.0], []]
```
or 
```
[[1, -12.516753754312305], [-6D9107941, 255.0], []]
```

m6w6:
The global decContext `g_IonEventDecimalContext` has traps set to
DEC_Errors, which wouldn't be set if `decContextDefault()` would be
used with any `kind` but DEC_INIT_BASE.

*Issue #, if available:* #221

*Description of changes:*
Set  `g_IonEventDecimalContext.traps` to 0 (zero).

### DIFF
--- a/tools/events/ion_event_util.cpp
+++ b/tools/events/ion_event_util.cpp
@@ -22,7 +22,7 @@ decContext g_IonEventDecimalContext = {
         DEC_MAX_MATH,                   // max exponent
         -DEC_MAX_MATH,                  // min exponent
         DEC_ROUND_HALF_EVEN,            // rounding mode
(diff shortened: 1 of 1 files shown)


tgregg (approved):
Good find. Thanks for the fix!

PR #262, opened 2021-12-06
fix leak with ion_writer_open_stream()

m6w6:
See line 145


tgregg (approved):
Good find, thanks for the fix.

PR #263, opened 2021-12-06
fixup #260 i.e. f7d33b0

m6w6:
*Issue #, if available:* #260

*Description of changes:*

The fix for #260 uses `p_options`, which might be `NULL`. 
Using `pwriter->options`, configured by any `p_options` instead.


cheqianh (approved):
Good catch thanks!

PR #265, opened 2021-12-08
Fix an issue where writer-owned resources were not being freed...

rgantt:
... When encountering errors while closing the writer

*Issue #, if available:* N/A

*Description of changes:* Use `UPDATEERROR` instead of `IONCHECK` so that the error encountered while flushing a too-small output buffer is propagated without preventing the subsequent cleanup work.

Note: running this test under `valgrind` before the change is applied shows one 64 KiB page definitely lost. After the change, it's freed properly.

### DIFF
--- a/test/test_ion_writer.cpp
+++ b/test/test_ion_writer.cpp
@@ -46,7 +46,7 @@ iERR ion_test_open_file_writer(hWRITER *writer, FILE *out, BOOL is_binary) {
     iRETURN;
 }
 
-TEST(WriterTest, ResourcesLeakedOnWriteToTooSmallBuffer)
+TEST_F(WriterTest, ResourcesNotLeakedOnWriteToTooSmallBuffer)
 {
     hWRITER writer = NULL;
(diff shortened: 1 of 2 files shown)


test/test_ion_writer.cpp
```diff
@@ -46,6 +46,23 @@ iERR ion_test_open_file_writer(hWRITER *writer, FILE *out, BOOL is_binary) {
     iRETURN;
 }
 
+TEST_F(WriterTest, ResourcesNotLeakedOnWriteToTooSmallBuffer)
+{
+    hWRITER writer = NULL;
+    ION_WRITER_OPTIONS options = {
+        .output_as_binary = true
```
tgregg, line 53: It looks like the Windows build (via AppVeyor) [doesn't like](https://ci.appveyor.com/project/tgregg/ion-c-3akm7/builds/41820078/job/3tdu1mqxms1th8ew#L397) this syntax.

tgregg (approved):
Thanks!

PR #268, opened 2021-12-17
fix wrong ASSERT assumption

m6w6:
*Description of changes:*
Assertion checks for non-equality against NULL instead of equality.


cheqianh (approved):
Thanks for the change!

cheqianh:
Thanks for the contribution!

PR #274, opened 2022-01-06
Refactor the ION_SYMBOL_TABLE type to make it opaque

### LINKED ISSUE #264: Symbol and page pool leak memory when used in a multi-threaded context
`ion-c` used to allocate a [global symbol table](https://github.com/amzn/ion-c/blob/a9af6be3f2ed38d4775257adfa5672c845cac361/ionc/ion_symbol_table.c#L215) that's used to pool/cache global symbols. This was later put into [thread-local storage (TLS)](https://github.com/amzn/ion-c/commit/44b6d5e688a1147fce1b74b616c640cfd12a6a17#diff-f5b0a06ee94bca57b08db344a393a3424ba865e145a73a3438c149ff38eb83d3R215), presumably because of thread safety concerns, and this was a cheap/minimally invasive mitigation until further work could be done to make things properly thread safe. Since the symbol pool now lives in TLS, each thread gets its own (global) symbol pool. When the symbol pool is initialized, allocations for the (growable) array backing the symbol pool are handed out from the heap (i.e. not from TLS itself - which wouldn't really make sense anyways since TLS is limited in size). Since the symbol pool is treated as a global resource, there isn't any code to free this backing heap storage - it's assumed to have the same lifetime as the program, which is a fine assumption if it truly were a global resource, but isn't correct given that it is now thread local.

When a thread terminates, it's TLS gets free'd, but nothing is cleaning up the heap-allocated parts of the global symbol pool (because there isn't anything that frees the heap-allocated parts of the symbol pool - it was originally implemented as a global resource). So whenever a thread dies or is reinitialized, we effectively leak its copy of the global symbol pool.

Something similar seems to be happening for the page pool, as I have observed leaks that go away when the page pool is disabled (i.e. the [default page limit](https://github.com/amzn/ion-c/blob/master/ionc/ion_alloc.h#L156) is set to 0) and it looks like [a pointer to the page pool is also stored in TLS](https://github.com/amzn/ion-c/blob/799c4d46520659a17b3fb74cd75a52e8277e4b8f/ionc/ion_alloc.h#L161).  Overall, about ~1.1 MB are being leaked per thread.  With lots of threads or a thread pool that recycles threads and re-initializes TLS though, this adds up pretty quickly.

sadderchris:
*Description of changes:*
Internally, ion-c pokes at the ION_SYMBOL_TABLE type quite a bit, which violates encapsulation and makes it hard to make changes to the data structure without breakage across the library.  This cleans up direct usages of the symbol table type, replaces them with API calls, and makes the ION_SYMBOL_TABLE type opaque to avoid depending on the internal details outside of ion_symbol_table.c.

This is needed in preparation to fix #264, as we need to make some changes to the representation of the ION_SYMBOL_TABLE type to fix it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ message(STATUS "Setting DECNUMBER max digits to 34")
 
 set(CMAKE_INSTALL_RPATH "$ORIGIN")
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 add_subdirectory(decNumber)
 add_subdirectory(ionc)
--- a/ionc/ion_catalog.c
+++ b/ionc/ion_catalog.c
@@ -236,7 +236,7 @@ iERR _ion_catalog_find_symbol_table_helper(ION_CATALOG *pcatalog, ION_STRING *na
     iRETURN;
 }
 
-iERR ion_catalog_find_best_match(hCATALOG hcatalog, iSTRING name, long version, ION_SYMBOL_TABLE **p_psymtab)
+iERR ion_catalog_find_best_match(hCATALOG hcatalog, iSTRING name, long version, hSYMTAB *p_symtab)
 {
     iENTER;
     ION_CATALOG      *catalog;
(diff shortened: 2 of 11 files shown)


tgregg:
Note: ion-test-driver failed for the following reason: "Read_error: new commit has different read error(s)." (Full details below.) This is fine, as both before and after the change, ion-c failed with IERR_INVALID_SYMBOL for this input. The test driver is complaining because the failure occurred in different places, which is overly strict. Once reviewed and approved, we'll merge this change even though ion-test-driver fails for this reason.

```
$ion_1_0
{
 'subfieldVarUInt32bit.ion': {
  read_error: {
   message: "Read_error: new commit has different read error(s).",
   read_error: {
    'ion-c_9310a99': ErrorReport::[
     {
      error_type: READ,
      message: "/Users/runner/work/ion-c/ion-c/output/build/ion-c/tools/events/ion_event_stream.cpp:328 : IERR_INVALID_SYMBOL",
      location: "/Users/runner/work/ion-c/ion-c/output/build/ion-tests_ef0451a/iontestdata/good/subfieldVarUInt32bit.ion"
     }
    ],
    'ion-c_677ddfd': ErrorReport::[
     {
      error_type: READ,
      message: "/Users/runner/work/ion-c/ion-c/output/build/ion-c/tools/events/ion_event_stream.cpp:1062 : IERR_INVALID_SYMBOL",
      location: "/Users/runner/work/ion-c/ion-c/output/build/ion-tests_ef0451a/iontestdata/good/subfieldVarUInt32bit.ion"
     }
    ]
   }
  }
 }
}
```

tgregg (approved):
Looks good.

PR #275, opened 2022-01-12
ensure we pass the cloned symbol table's owner to the helpers

m6w6:
*Issue #, if available:* -

*Description of changes:*
If `ion_symbol_table_clone_with_owner` is called with `owner==NULL`, i.e. to make itself its owner, `NULL` is still passed through to helper functions, which should receive `clone->owner` instead.


tgregg (approved):
Thanks for fixing this!

PR #278, opened 2022-01-26
Adds ion_reader_get_value_position for fetching the line & column of the current value from a text reader.

dlurton:
The position of the current value is the first character of its Ion-text representation or (if present) the first character of the first annotation.  This information is useful for reporting semantic errors to end-users when Ion-text files are used to express domain-specific language (DSL) scripts.

Also fixes a bug causing the current line number from being set to zero after every call to `ion_reader_next`.

### DIFF
--- a/ionc/include/ionc/ion_reader.h
+++ b/ionc/include/ionc/ion_reader.h
@@ -316,6 +316,19 @@ ION_API_EXPORT iERR ion_reader_get_value_offset    (hREADER   hreader
  * to end-users when Ion-text files are used to express domain-specific
  * language (DSL) scripts.
  *
+ * The numbers reported by this function (p_line) start at 1, while the
+ * column offset (p_col_offset) start at zero (since it is the
+ * offset from the start of the line.
+ *
+ * If the last call to `ion_reader_next` encountered a container
+ * terminator (`]`, `)` or `}`), instead of another value, the position
+ * reported is that of the terminator.  Thus, it is possible to obtain
+ * the starting and ending positions of all containers.
+ *
+ * If the last call to `ion_reader_next` encountered a the end of file
+ * instead of another value, the position reported is that of the last
+ * character in the file.
+ *
  * If the reader is a binary reader, fails immediately and returns
  * IERR_INVALID_ARG.  If all that's desired is the offset of the value
  * relative to the beginning of the buffer, please use
--- a/ionc/include/ionc/ion_reader.h
+++ b/ionc/include/ionc/ion_reader.h
@@ -317,7 +317,7 @@ ION_API_EXPORT iERR ion_reader_get_value_offset    (hREADER   hreader
  * language (DSL) scripts.
  *
  * The numbers reported by this function (p_line) start at 1, while the
- * column offset (p_col_offset) start at zero (since it is the
+ * column offset (p_col_offset) start at zero since it is the
  * offset from the start of the line.
  *
  * If the last call to `ion_reader_next` encountered a container
@@ -325,7 +325,7 @@ ION_API_EXPORT iERR ion_reader_get_value_offset    (hREADER   hreader
  * reported is that of the terminator.  Thus, it is possible to obtain
  * the starting and ending positions of all containers.
  *
- * If the last call to `ion_reader_next` encountered a the end of file
+ * If the last call to `ion_reader_next` encountered the end of file
  * instead of another value, the position reported is that of the last
  * character in the file.
  *
--- a/ionc/include/ionc/ion_reader.h
+++ b/ionc/include/ionc/ion_reader.h
@@ -443,12 +443,11 @@ ION_API_EXPORT iERR ion_reader_read_lob_bytes        (hREADER hreader, BYTE *p_b
 ION_API_EXPORT iERR ion_reader_read_lob_partial_bytes(hREADER hreader, BYTE *p_buf, SIZE buf_max, SIZE *p_length);
 
 /**
- * Gets the offset of from the current position and if hreader is a text
- * reader, also gets the line and column of the current position of the
- * reader.
+ * Gets the current position and if hreader is a text reader, also gets
+ * the line and column numbers.
  *
- * Note that this is only useful for error reporting debugging purposes
- * about invalid Ion data, since the position reported by this function
+ * Note that this is only useful for error reporting or debugging purposes
+ * about malformed Ion data, since the position reported by this function
  * is unlikely to be pointed at the start of a value.  To obtain the
  * position of the current value from the reader to report the origin
  * of semantic errors within well formed Ion data, see
--- a/ionc/include/ionc/ion_reader.h
+++ b/ionc/include/ionc/ion_reader.h
@@ -306,6 +306,27 @@ ION_API_EXPORT iERR ion_reader_set_symbol_table    (hREADER   hreader
  */
 ION_API_EXPORT iERR ion_reader_get_value_offset    (hREADER   hreader
                                                    ,POSITION *p_offset);
+
+/**
+ * Gets the position of the current value in the reader.
+ *
+ * The position of the current value is the first character of its Ion-text
+ * representation or (if present) the first character of the first
+ * annotation.  This information is useful for reporting semantic errors
+ * to end-users when Ion-text files are used to express domain-specific
+ * language (DSL) scripts.
+ *
+ * If the reader is a binary reader, fails immediately and returns
+ * IERR_INVALID_ARG.  If all that's desired is the offset of the value
+ * relative to the beginning of the buffer, please use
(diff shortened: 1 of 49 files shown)


test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
     ASSERT_EQ(-4, value);
 }
+
+/** tests ion_reader_get_value_position for scalar values. */
+TEST(IonTextPosition, PositionOfValues) {
+    const char *ion_text =
+        // Column number cheat sheet:
+        // 1         2         3         4         5         6         7         8         9         10        11        12        13        14
+        // 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
+        "  null true 1 2e0 3.0 four 'five' \"six\" 2007-07-07T {{ V2hhdCBhcmUgeW91IGRvaW5nLCBEYXZlPw== }} {{ '''I can't let you do that, Dave.''' }} 42";
```
tgregg, line 730: Maybe I'm misunderstanding how to use the cheat sheet, but I was trying to use it to easily look up offsets. The top number is the tens place and the bottom is the ones place. However, there are currently two problems: 1) (minor) the tens place number should start at 0, and 2) it's misaligned by two since the data starts with two spaces, which are included in the offset calculations. Still really helpful though. (The same goes for the cheat sheet for the next test.)

test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
+
+    int64_t offset = 0;
+    int32_t line = 0;
+    int32_t col_offset = 0;
+
+    ION_ASSERT_OK(ion_reader_next(reader, &type));
+    ASSERT_EQ(tid_LIST, type);
+    ION_ASSERT_OK(ion_reader_get_value_position(reader, &offset, &line, &col_offset));
+    ASSERT_EQ(offset, 1);
+    ASSERT_EQ(line, 2);
```
tgregg, line 912: So lines start at 1, while offsets start at 0. I think that's fine, but it would be good to make sure it's documented on the function.

test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
+    // third list element
+    ION_ASSERT_OK(ion_reader_next(reader, &type));
+    ASSERT_EQ(tid_INT, type);
+    ION_ASSERT_OK(ion_reader_get_value_position(reader, &offset, &line, &col_offset));
+    ASSERT_EQ(offset, 15);
+    ASSERT_EQ(line, 5);
+    ASSERT_EQ(col_offset, 3);
+
+    // finally, step out
+    ION_ASSERT_OK(ion_reader_step_out(reader));
```
tgregg, line 948: What happens if you do another next() before stepping out to find the end of the container? Would be good to add verification here. (This is common when looping through all of a container's elements.)

test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
+    ION_ASSERT_OK(ion_reader_get_value_position(reader, &offset, &line, &col_offset));
+    ASSERT_EQ(offset, 20);
+    ASSERT_EQ(line, 7);
+    ASSERT_EQ(col_offset, 1);
+}
+
+/** tests ion_reader_get_value_position for sexp elements appearing on separate lines. */
+TEST(IonTextPosition, PositionOfSexpElements) {
+    // this is `(1 2 3) 42` split across multiple lines with values prefixed with various tabs and spaces.
+    const char *ion_text = "\n(\n\t1 \n\t 2 \n\t  3\n)\n 42";
```
tgregg, line 962: Would be good to annotate one of them.

test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
+    ION_ASSERT_OK(ion_reader_get_value_position(reader, &offset, &line, &col_offset));
+    ASSERT_EQ(offset, 20);
+    ASSERT_EQ(line, 7);
+    ASSERT_EQ(col_offset, 1);
+}
+
+/** ion_reader_get_value_position for struct fields appearing on separate lines. */
+TEST(IonTextPosition, PositionOfStructFields) {
+    // this is `{ a: 1, b: 2, c: 3 } 42` split across multiple lines with fields prefixed with various whitespace.
+    const char *ion_text = "\n{a:1,\nb: 2,\nc:  \n3\n} 42";
```
tgregg, line 1027: Let's annotate one of the field values.

test/test_ion_text.cpp
```diff
@@ -720,3 +720,426 @@ TEST(IonTextInt, BinaryLiterals) {
...
+    // Advance to "4", read it and assert value and position.
+    ION_ASSERT_OK(ion_reader_next(reader, &type));
+    ION_ASSERT_OK(ion_reader_read_int32(reader, &int32));
+    ASSERT_EQ(4, int32);
+    // Read the position of the int and verify we have the correct position
+    ION_ASSERT_OK(ion_reader_get_value_position(reader, &offset, &line, &col_offset));
+    ASSERT_EQ(offset, 7);
+    ASSERT_EQ(line, 5);
+    ASSERT_EQ(col_offset, 0);
+}
```
tgregg, line 1145: Another interesting test case may be to make sure the offsets work properly with long strings, which may have newlines and/or comments between fragments. E.g.

```
'''foo'''
// bar
'''baz'''
```
which is equivalent to `"foobaz"`.

tgregg (commented):
Looks good. Just a couple minor comments and test suggestions.

ionc/ion_reader.c
```diff
@@ -2156,6 +2155,42 @@ iERR ion_reader_get_value_offset(hREADER hreader, POSITION *p_offset)
     iRETURN;
 }
 
+iERR ion_reader_get_value_position(hREADER hreader, int64_t *p_offset, int32_t *p_line, int32_t *p_col_offset) {
+    iENTER;
+    ION_READER *preader;
+    POSITION    offset;
+    int32_t     line = -1;
+    int32_t     col_offset = -1;
```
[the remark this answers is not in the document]
dlurton: This follows a pattern I saw repeated elsewhere in the same file, i.e. in the `ion_reader_get_value_offset` immediately above.  This seems to avoid modifying the output parameters at all unless the function is going to succeed.  That said, since these are just being passed down to `_ion_reader_text_get_value_position` which does the same thing, it's not really necessary in this case.  I'll do as you suggest.

ionc/include/ionc/ion_reader.h
```diff
@@ -306,6 +306,27 @@ ION_API_EXPORT iERR ion_reader_set_symbol_table    (hREADER   hreader
...
+ * representation or (if present) the first character of the first
+ * annotation.  This information is useful for reporting semantic errors
+ * to end-users when Ion-text files are used to express domain-specific
+ * language (DSL) scripts.
+ *
+ * If the reader is a binary reader, fails immediately and returns
+ * IERR_INVALID_ARG.  If all that's desired is the offset of the value
+ * relative to the beginning of the buffer, please use
+ * ion_reader_get_value_offset, which works with both text and binary
+ * readers.
```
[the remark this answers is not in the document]
tgregg: Yeah, there isn't a place for text-specific APIs. If we add one in the future, we could deprecate this. Or, we could eventually make `p_offset` useful for binary.

tgregg (commented):
I resolved the comments that are addressed; I think my comments on the tests are still actionable.

test/test_ion_text.cpp
dlurton: Yeah, fixing this.  Also removing the other, less needed one where I check for struct field positions.

test/test_ion_text.cpp
dlurton: added.

test/test_ion_text.cpp
dlurton: happens to be positioned on the closing `]`, `)` or `}`, which is convenient.  added tests and documentation.

dlurton: also added a similar call to test the position after `ion_reader_next()` encounters a legit EOF.

test/test_ion_text.cpp
dlurton: I was certain there'd be a bug here, but in fact, there wasn't.

dlurton:
> I resolved the comments that are addressed; I think my comments on the tests are still actionable.

Yes, was still working on this.  Have another push coming in a minute.

tgregg (approved):
Looks good!

tgregg (approved):

PR #280, opened 2022-02-02
Delete the page pool allocator

sadderchris:
*Issue #, if available:* #264

*Description of changes:*
The page pool allocator leaks memory when used in a threaded context (#264). While we could invest in making this thread-safe by using proper concurrency primitives, it's not clear that the page pool allocator is actually providing any benefit right now.

There are no benchmarks to indicate that it is, and any sane allocator should already be doing something like this behind the scenes (or should be configurable to, in a thread-safe way to boot).  I ran some (limited) benchmarks on the `ionizer` cli, and there wasn't a significant difference in performance without the page pool (the timings were within margin of error, a handful of microseconds at most - not noticeable on the human scale, at least for `ionizer`'s use cases).

This nukes the page pool code.  We can perform some more perf tests and benchmarks to determine if it's really worth it to reintroduce something like it in the future.

### DIFF
--- a/ionc/ion_alloc.h
+++ b/ionc/ion_alloc.h
@@ -131,48 +131,6 @@ typedef struct _ion_allocation_chain DBG_ION_ALLOCATION_CHAIN;
 #endif
 
 
-//
-//  structures and functions for the ion allocation page pool
-//  this is a list of free pages which are used by the various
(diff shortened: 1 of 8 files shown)


tgregg (approved):
Looks good, thank you.

PR #293, opened 2022-04-05
fix #292: bad alloc performance due to excessive memset

### LINKED ISSUE #292: bad performance due to zeroing large buffers
I just stumbled over a `memset` in `ion_allocation.c:172`, which seems to cause a quite big performance hit. 

Consider the following two callgraphs of the same code, just once with that `memset` commented out:
## With memset:
binary-w_memset

## Without memset:
binary-wo_memset

m6w6:
*Issue #, if available:* #292

*Description of changes:*
* remove `memset` in `_ion_alloc_with_owner_helper`
* fix crashes and failing tests


tgregg (approved):
Looks good to me. Thanks, @m6w6 !

PR #295, opened 2022-04-25
fix broken clob parsing when scanning unknown lobs

sadderchris:
*Issue #, if available:* N/A

*Description of changes:*
Skip-scanning through `[{{"foo"}},{{"bar"}}]` in text mode currently (erroneously) fails with `IERR_UNEXPECTED_EOF`.

`_ion_scanner_skip_unknown_lob` (which handles skipping clobs and blobs when skip-scanning through a data structure) expects to parse the closing `}}` of a clob or blob.  However, when it tries to skip a "plain clob" (a clob with a double-quoted string), it calls `_ion_scanner_skip_plain_clob`, which consumes the closing `}}`.  So it erroneously continues reading and skipping input, even though it's already skipped the entire clob, leading to the `IERR_UNEXPECTED_EOF` result.

This change identifies the unknown lob by the matching against the first non-whitespace character it sees and reuses the helper functions for skipping the particular lob type of the unknown lob (clob with long strings, plain clob, or blob) instead of duplicating the skipping logic from these functions into `_ion_scanner_skip_unknown_lob`.

### DIFF
--- a/ionc/ion_scanner.c
+++ b/ionc/ion_scanner.c
@@ -1184,30 +1184,21 @@ iERR _ion_scanner_skip_unknown_lob(ION_SCANNER *scanner)
     iENTER;
     int c;
 
-    for (;;) {
-        // we should see a double closing curly brace next
-        IONCHECK(_ion_scanner_read_past_lob_whitespace(scanner, &c));
-        switch (c) {
-        case '\'':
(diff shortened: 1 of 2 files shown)


waffleking32 (approved):
Looks good. Thanks!

tgregg (approved):
Thanks for the fix.

PR #296, opened 2022-05-30
fix _ion_writer_add_annotation_helper memory accumulation

soospeter:


tgregg (commented):
Thanks for the contribution!

Although this is technically a behavioral change (before, the user could safely free the provided ION_STRING immediately after calling this function; now, that is unsafe), I'm okay with it as I think it's unlikely anyone would be doing that. It also follows the same pattern as `ion_writer_write_field_name*`.

One request: could you please add a comment to this API stating this requirement. See the comment on [`ion_writer_write_field_name`](https://github.com/amzn/ion-c/blob/master/ionc/include/ionc/ion_writer.h#L218), for example.

tgregg (approved):

PR #301, opened 2022-07-09
Fix reader, stream, and import leaks in CLI 

nirosys:
*Issue #, if available:* None

*Description of changes:*
Minor fixes to release memory in two spots within cli.cpp.

The first was when opening a writer via `ion_cli_open_writer` with imports, the `has_imports` flag would not be set on the writer context, resulting in leaked imports from `ion_event_writer_close` not closing the imports.

The second is a stream and reader leak when cleaning up the reader_contexts in `ion_cli_command_compare_standard`.

---


cheqianh (approved):
LGTM!

PR #302, opened 2022-07-09
Fix leak of resources when unable to flush binary writer

nirosys:
*Issue #, if available:* None

*Description of changes:*
Prior to this commit when a binary writer was closed,
but could not be flushed, the `_ion_writer_close_helper`
function would skip `_ion_writer_binary_close`, which
resulted in the binary writer's `_value_stream` being
leaked.

This commit updates the close functions for both binary
and text formats to include a BOOL to communicate if
we want to flush or not. `_ion_writer_close_helper` was then
updated to track if we are able to flush and provide
that info to the format writer's close, so we can ensure
all resources are cleaned up.

---

### DIFF
--- a/ionc/ion_writer.c
+++ b/ionc/ion_writer.c
@@ -2829,9 +2829,12 @@ iERR _ion_writer_close_helper(ION_WRITER *pwriter)
 
     ASSERT(pwriter);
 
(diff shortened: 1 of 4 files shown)


tgregg (approved):
Looks good

PR #303, opened 2022-07-09
Fix memleak when writer is closed prior to closing containers

nirosys:
*Issue #, if available:* None

*Description of changes:*
ION_WRITER's `_pending_temp_entity_pool` is moved over to `_temp_entity_pool` while transitioning from the symtab intercept state, when depth is zero.

If the writer was closed prior to that transition, the data pointed to by `_pending_temp_entity_pool` would be leaked. This PR adds a function to free the data, and checks during `_ion_writer_free` to ensure the pending temp entity pool is freed if needed, similar to how the temp entity pool is freed.

---

### DIFF
--- a/ionc/ion_writer.c
+++ b/ionc/ion_writer.c
@@ -2812,6 +2812,7 @@ iERR _ion_writer_free(ION_WRITER *pwriter)
     // Free the writer's members.
     UPDATEERROR(_ion_writer_free_local_symbol_table( pwriter ));
     UPDATEERROR( _ion_writer_free_temp_pool( pwriter ));
+    UPDATEERROR(_ion_writer_free_pending_pool(pwriter));
 
     // If the writer allocated the stream, free it.
     if (pwriter->writer_owns_stream) {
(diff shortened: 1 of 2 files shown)


ionc/ion_writer.c
```diff
@@ -3281,3 +3282,14 @@ iERR _ion_writer_free_temp_pool( ION_WRITER *pwriter )
...
+
+iERR _ion_writer_free_pending_pool(ION_WRITER *writer) {
+    iENTER;
+
+    if (writer->_pending_temp_entity_pool != NULL) {
+        ion_free_owner(writer->_pending_temp_entity_pool);
+        writer->_pending_temp_entity_pool = NULL;
+    }
+
+    RETURN(__location_name__, __line__, __count__++, err);
```
tgregg, line 3294: Is this an intentional choice not to use `iRETURN`?

Should this just be a `void` function since it looks like it can't return anything but `IERR_OK`?

nirosys: Yea that was intentional. Since the `fail` label that `iRETURN` defines isn't used, using it over `RETURN` only adds to the warnings generated at build.

It could definitely be a `void` function. I maintained the signature that the other free functions had for writer members, but all of them could be converted to void since all 3 only return `IERR_OK`.

tgregg: Got it. So maybe we need a separate (and larger) task to get rid of unnecessary `iRETURN` usages, if we want to reduce warnings. Approving this PR.

tgregg (approved):

PR #304, opened 2022-07-09
Fix a couple leaks around event and symbol cleanup

nirosys:
*Issue #, if available:* None

*Description of changes:*
This PR addresses two instances in ion_event_stream where leaks were occurring.

The first is when cleaning up the annotation symbols, and field name, at the end of `ion_event_stream_read_event`. The string data for `value_field_name`, as well as each of the symbols stored in `value_annotations` was not freed.

The second was when removing an event from the event stream in `ion_event_stream_is_event_stream`. Once the pointer to the event is removed from the vector, the object itself needed to be freed.

---


cheqianh:
test-driver failed due to different line number, 
`ion_event_stream.cpp:1063 : IERR_INVALID_SYMBOL` vs. `ion_event_stream.cpp:1076 : IERR_INVALID_SYMBOL`
which is correct and expected.

PR #305, opened 2022-07-09
Fix leak of reader during path extraction

nirosys:
*Issue #, if available:* None

*Description of changes:*
During the execution of `ion_extractor_path_create_from_ion`, a reader is created that does not get freed.

I added the freeing of the reader, but also removed the use of `iRETURN` and manually defined the fail label, so that I could ensure all paths lead to the reader being closed.

---


tgregg (approved):
Thanks, this one was definitely my fault.

PR #310, opened 2022-12-01
Add support for down-converting ion to JSON

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR adds functionality to the ion text writer to allow for a basic lossy conversion to JSON encoding. The down-conversion is done in the same manner described in the [cookbook](https://amzn.github.io/ion-docs/guides/cookbook.html#down-converting-to-json), with a couple minor differences that will be corrected soon.

The most noticeable difference is the ranges in which unicode is escaped. Currently, any non-ascii characters are emitted as unicode escape sequences. I'll follow up with a fix for that soon. I have identified the issue, and just need to implement the fix and test it.

An existing issue that needs some consensus is how multiple top-level values are handled. With this commit multiple top-level items are serialized with no delimeter. This is supported by some JSON parsers, such the one used by the tool jq, but may be an issue with others. A simple hacky fix can be added to place all top-level values into a list, but I didn't feel comfortable with the implementation unless it is truly needed. Another option might be to use ~[JSON Lines](https://jsonlines.org/)~ (website seems to be down).

The changes add a new option to `ION_WRITER_OPTIONS` called `json_downconvert` which, when true will format any output written to the writer using the JSON friendly formatting described in the cookbook link (minus the above notes).

An example of converting ion data to json, can be found in `test/test_ion_text.cpp` in the `convert_to_json` function.

This PR also extends the CLI to allow the output formats `json` and `json-pretty`, which perform the down conversion with pretty printing off, and on, respectively. Examples of using `ion process` to perform downconversion has also been added to the help.

---


nirosys:
MacOS issue was due to a bug in Xcode 14.0.1. I'm running 14.1 and I was unable to reproduce the issue, so I've added a build step for macos builds in the GH Actions workflow to set the default xcode to 14.1 (it is installed in the image, just not used by default). That seems to have fixed the MacOS issue.. now on to the windows issue..

nirosys:
Reproduced the issue in the windows build, tracked it down to a misplaced `#endif`. Everything should be green now.

ionc/ion_writer_text.c
```diff
@@ -667,7 +730,11 @@ iERR _ion_writer_text_write_timestamp(ION_WRITER *pwriter, iTIMESTAMP value)
         temp[output_length] = '\0';
 
         // and our helper does the rest
+        if (json_downconvert)
+           ION_PUT(pwriter->output, '"');
         IONCHECK(_ion_writer_text_append_ascii_cstr(pwriter->output, temp));
+        if (json_downconvert)
+           ION_PUT(pwriter->output, '"');
```
tgregg, line 737: We haven't defined a formal style guide for this repo, but let's use braces for every `if` (for all new code anyway).

ionc/ion_writer_text.c
```diff
@@ -679,28 +746,30 @@ iERR _ion_writer_text_write_symbol_from_string(ION_WRITER *pwriter, ION_STRING *
...
     SIZE written;
+    BOOL down_convert = ION_TEXT_WRITER_IS_JSON();
 
     if (pwriter->depth == 0 && pwriter->annotation_count == 0 && pstr->value[0] == '$'
         && _ion_symbol_table_parse_version_marker(pstr, NULL, NULL)) {
         // The text $ion_<int>_<int> is reserved for the IVMs. This is a no-op.
         SUCCEED();
     }
     else {
+        char quote = ION_TEXT_WRITER_IS_JSON() ? '"':'\'';
```
tgregg, line 757: Can use `down_convert` here for consistency.

ionc/ion_writer_text.c
```diff
@@ -1128,24 +1204,26 @@ iERR _ion_writer_text_close(ION_WRITER *pwriter, BOOL flush)
     iRETURN;
 }
 
-iERR _ion_writer_text_append_symbol_string(ION_STREAM *poutput, ION_STRING *p_str, BOOL as_ascii, BOOL system_identifiers_need_quotes)
+iERR _ion_writer_text_append_symbol_string(ION_WRITER *pwriter, ION_STRING *p_str, BOOL system_identifiers_need_quotes)
 {
     iENTER;
     SIZE written;
+    char quote_char = (pwriter->options.json_downconvert) ? '"' : '\'';
```
tgregg, line 1211: `ION_TEXT_WRITER_IS_JSON()` for consistency

ionc/ion_writer_text.c
```diff
@@ -1128,24 +1204,26 @@ iERR _ion_writer_text_close(ION_WRITER *pwriter, BOOL flush)
...
 
     if (!poutput) FAILWITH(IERR_BAD_HANDLE);
     if (!p_str) FAILWITH(IERR_INVALID_ARG);
     if (p_str->length < 0) FAILWITH(IERR_INVALID_ARG);
 
-    if (_ion_symbol_needs_quotes(p_str, system_identifiers_need_quotes)) {
-        ION_PUT(poutput, '\'');
-        if (as_ascii) {
-            IONCHECK(_ion_writer_text_append_escaped_string(poutput, p_str, '\''));
+    if (_ion_symbol_needs_quotes(p_str, system_identifiers_need_quotes) || pwriter->options.json_downconvert) {
```
tgregg, line 1218: `ION_TEXT_WRITER_IS_JSON()` for consistency

ionc/ion_writer_text.c
```diff
@@ -1181,7 +1259,10 @@ iERR _ion_writer_text_append_escape_sequence_string(ION_STREAM *poutput, BYTE *c
 
     c = *cp;
     if (c < 32 || c == '\\' || c == '"' || c == '\'') {
-        image = _ion_writer_get_control_escape_string(c);
+        if (!down_convert)
+            image = _ion_writer_get_control_escape_string(c);
+        else
+            image = _ion_writer_get_control_escape_string_json(c);
```
tgregg, line 1265: I have a minor preference to swap the bodies and lose the negation. Disregard if this style was chosen deliberately.

test/test_ion_text.cpp
```diff
@@ -1332,3 +1331,137 @@ TEST(IonTextPosition, LongStringPositions) {
...
+    // Integers
+    IONJSON_CMP("123456789012345678901", "123456789012345678901");
+
+    // Floats
+    IONJSON_CMP("nan", "null");
+    IONJSON_CMP("+inf", "null");
+    IONJSON_CMP("-inf", "null");
+    IONJSON_CMP("1.0e0", "1");
+    IONJSON_CMP("1.5e0", "1.5");
+    IONJSON_CMP("1e-5", "1e-05");
```
tgregg, line 1410: That's... interesting? Is `05` necessary?

test/test_ion_text.cpp
```diff
@@ -1332,3 +1331,137 @@ TEST(IonTextPosition, LongStringPositions) {
...
+    IONJSON_CMP("0.1",  "0.1");
+
+    // Decimals
+    // Decimal formatting does not follow the same pattern as float formatting.
+    IONJSON_CMP("1d0", "1");
+    IONJSON_CMP("1.",  "1");
+    IONJSON_CMP("1d-0",  "1");
+    IONJSON_CMP("0d1", "0E+1");
```
tgregg, line 1418: Looks like we're using lowercase `e` for some, and uppercase `E` for others?

tgregg (approved):
Nice. Minor comments inline.

Regarding how multiple top-level values should be handled, I'm fine with the behavior as-is. Namely, that newlines will separate top-level values when `pretty` is enabled, but not when `pretty` is disabled. If users need to use the output with a JSON parser that doesn't support multiple top-level values, they can easily accommodate that manually, e.g. by wrapping everything in a single list.

We could add a `lines` format option, but I see that as orthogonal to this change, as it could apply to both text Ion and JSON. I'm aware of use cases where "Ion lines" has come in handy.

ionc/ion_writer_text.c
nirosys: Yea, that wasn't done for any particular reason, definitely reads better without the negation.

test/test_ion_text.cpp
nirosys: The exponent of 5 is needed to force the float to get rendered in exponential notation. The literal `05` in the string isn't required by anything, it's just the way ion-c is currently rendering exponents to text.

test/test_ion_text.cpp
nirosys: Yea, the way decimals are rendered differs from how floats are. The decNumber library handles the decimals, while we're using `snprintf` for floats. I'm adding a task to get those following the same rules to the plan. Negligible priority I'd imagine, but annoying none the less.

PR #316, opened 2023-01-23
Add missing macro usage to allow building with DEBUG defined

nirosys:
*Issue #, if available:* N/A

*Description of changes:*
The macro ecosystem that ion-c defines, includes the ability to get some diagnostic information when building with `DEBUG` defined. Currently neither the Debug, nor the Release build actual use `DEBUG`. The diagnostics provided may be useful for debugging issues in ion-c or ion-python. Since the current builds do not include the debug macros there were a couple of areas where variables like `__file__`, and `__line__` where not being defined due to a mismatch between failure macros, and enter macros.

This PR addresses those areas and corrects the build when DEBUG is defined.

Here is an example of the diagnostic messages when running unit tests:
```
[       OK ] IonTimestampOutOfRangeFractionParameterized/IonTimestampOutOfRangeFraction.WriterFailsOnOutOfRangeFraction/5 (0 ms)
[ RUN      ] IonTimestampOutOfRangeFractionParameterized/IonTimestampOutOfRangeFraction.WriterFailsOnOutOfRangeFraction/6

ERROR 10 [IERR_INVALID_TIMESTAMP] WITH MESSAGE 'Invalid fraction value.' AT LINE 176 IN ion_timestamp.c

ERROR 10 [IERR_INVALID_TIMESTAMP] WITH MESSAGE 'fraction seconds can't be greater than 1' AT LINE 854 IN ion_timestamp.c
[       OK ] IonTimestampOutOfRangeFractionParameterized/IonTimestampOutOfRangeFraction.WriterFailsOnOutOfRangeFraction/6 (0 ms)
[----------] 7 tests from IonTimestampOutOfRangeFractionParameterized/IonTimestampOutOfRangeFraction (1 ms total)
```

---

### DIFF
--- a/ionc/ion_writer_text.c
+++ b/ionc/ion_writer_text.c
@@ -606,7 +606,7 @@ iERR _ion_writer_text_write_double(ION_WRITER *pwriter, double value)
 }
 
 iERR _ion_writer_text_write_double_json(ION_WRITER *pwriter, double value) {
-   iERR err = IERR_OK;
+   iENTER;
    char image[64], *mark;
    int fpc = FLOAT_CLASS(value);
 
--- a/tools/cli/cli.cpp
+++ b/tools/cli/cli.cpp
@@ -61,6 +61,7 @@ iERR ion_cli_command_process_traverse(IonEventWriterContext *writer_context, Ion
 }
 
 iERR ion_cli_close_reader(IonCliReaderContext *context, iERR err, IonEventResult *result) {
+    FN_DEF;
     ASSERT(context);
     ION_SET_ERROR_CONTEXT(&context->input_location, NULL);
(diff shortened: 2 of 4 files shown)


tgregg (approved):
Nice.

I think there are two separate issues with the ion-test-driver run, causing it to fail. One is that files are not found. My guess is that this is due to the URL change that came from the organization rename. I opened https://github.com/amazon-ion/ion-test-driver/pull/27 to address that (and it looks like there are separate issue(s) with that). The other is that there are tons of failures simply due to line numbers changing. I've opened https://github.com/amazon-ion/ion-c/issues/317 to look into addressing that.

This PR doesn't need to wait on resolution of those issues.

Issue #318, opened 2023-01-27
Failure to seek when using user managed streams.


nirosys:
There is what seems to be a bug with seeking, when using a user controlled stream. The issue occurs when the seek position is more than 1 read away from the symbol table.

It appears that a call to `ion_reader_seek` will only result in a single call to the stream handler. The seek target is then used to update the stream position after that single read, which puts the reader in a bad state, presumably looking at bad memory.

---

The data I'm using for repro:
```C++
// echo '{hello: "World"}{dest1: "here"}{dest2: "now here"}{dest3: "here too"}' | ion dump --format binary | xxd -i
uint8_t ION_DATA[] = {
  /* 00 */ 0xe0, 0x01, 0x00, 0xea,                                           // Ion 1.0 Version Marker
  /* 04 */ 0xee, 0xa2, 0x81, 0x83,                                           // '$ion_symbol_table'::
  /* 08 */ 0xde, 0x9e,                                                       // {
  /* 10 */ 0x86,                                                             //   'imports':
  /* 11 */ 0x71, 0x03,                                                       //      $ion_symbol_table,
  /* 13 */ 0x87,                                                             //      'symbols':
  /* 14 */ 0xbe, 0x98,                                                       //        [
  /* 16 */ 0x85, 0x68, 0x65, 0x6c, 0x6c, 0x6f,                               //          "hello",
  /* 22 */ 0x85, 0x64, 0x65, 0x73, 0x74, 0x31,                               //          "dest1",
  /* 28 */ 0x85, 0x64, 0x65, 0x73, 0x74, 0x32,                               //          "dest2",
  /* 34 */ 0x85, 0x64, 0x65, 0x73, 0x74, 0x33,                               //          "dest3",
                                                                             //        ]
                                                                             // }
  /* 40 */ 0xd7, 0x8a, 0x85, 0x57, 0x6f, 0x72, 0x6c, 0x64,                   // {'hello': "World"}
  /* 48 */ 0xd6, 0x8b, 0x84, 0x68, 0x65, 0x72, 0x65,                         // {'dest1': "here"}
  /* 55 */ 0xda, 0x8c, 0x88, 0x6e, 0x6f, 0x77, 0x20, 0x68, 0x65, 0x72, 0x65, // {'dest2': "now here"}
  /* 66 */ 0xda, 0x8d, 0x88, 0x68, 0x65, 0x72, 0x65, 0x20, 0x74, 0x6f, 0x6f, // {'dest3': "here too"}
[...]

PR #320, opened 2023-02-20
Fix user stream reading to carry out multiple handler calls when needed

nirosys:
*Issue #, if available:* 318

*Description of changes:*
Prior to this PR when a user managed buffer is read with the intention of filling a page, the handler for the buffer is called only once. No information is provided to the handler in order to communicate how much data is needed.

This resulted in user managed buffers not correctly seeking and the reader falling out of sync with the underlying buffer. More details can be seen in issue #318.

This PR updates the `_ion_stream_fread` function to ensure that the user managed buffer handler is called enough times to fill the requested buffer, stopping early in the cases of error, or insufficient data.

---

### DIFF
--- a/test/test_ion_stream.cpp
+++ b/test/test_ion_stream.cpp
@@ -12,6 +12,7 @@
  * language governing permissions and limitations under the License.
  */
 
+#include <math.h>
 #include <gtest/gtest.h>
 #include "ion_event_util.h"
 #include <ionc/ion_types.h>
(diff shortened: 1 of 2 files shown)


test/test_ion_stream.cpp
```diff
@@ -99,7 +99,7 @@ TEST(IonStream, ContinuesOverPageBoundary) {
     ION_ASSERT_OK(ion_reader_close(reader));
     ION_ASSERT_OK(ion_stream_close(stream));
 
-    ASSERT_EQ(2, context.number_of_handler_invocations);
+    ASSERT_EQ(3, context.number_of_handler_invocations);
```
nirosys, line 102: The number of invocations changes since there is 1 more call now to establish that no data is left.

tgregg (approved):
Nice investigation, and nice fix.

Could you extract your repro from https://github.com/amazon-ion/ion-c/issues/318 into a test?

nirosys:
ty Tyler! Updated with a new commit adding a unit test based on the test program I had been used.

tgregg (approved):

PR #323, opened 2023-03-27
Bug Fix: User managed stream data corruption.

nirosys:
*Issue #, if available:* N/A

*Description of changes:*
This PR fixes two issues that were introduced with the recent fix for user managed streams.

The first is during the 'catch up' portion of `_ion_stream_fread`, where previously read data was copied into a new page. With the previous fix, the page's current pointer `dst` was not updated after the write causing subsequent reads to overwrite the data.

The second issue that was introduced is when a user managed stream handler returned an error, the error would result in zero bytes read, but no error was bubbled up. This would result in readers EOFing on error rather than identifying a read error.

This PR also adds better guards around data copies. I've noticed some handlers that set the buffer limit to NULL to signal data was not read. I think we need some more documentation on what is expected for the handlers.

---

### DIFF
--- a/ionc/ion_stream.c
+++ b/ionc/ion_stream.c
@@ -1824,8 +1824,10 @@ iERR _ion_stream_fread( ION_STREAM *stream, BYTE *dst, BYTE *end, SIZE *p_bytes_
                 } else { // No new data, so we break out.
                     break;
                 }
+            } else if (err != IERR_EOF) {
+                bytes_read = READ_ERROR_LENGTH;
+                break;
             } else {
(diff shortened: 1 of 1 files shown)


nirosys:
>_< I had added the fix for bubbling up the read error while looking at an ion-python issue, and forgot this morning that I hadn't run it against the unit tests. The change incorrectly also bubbled up an error when IERR_OK was returned, with null curr and/or limit pointers. Pushed a commit to address that.

tgregg (approved):

cheqianh (approved):

PR #325, opened 2023-05-09
Add version number to decNumber to ensure soft link creation

ruhland:
*Description of changes:*
Without a version the .so file is created directly without a link which will look something like this: `ionc/libionc.so -> libionc.so.1.1.2`
The missing link causes problems when building with yocto

### DIFF
--- a/decNumber/CMakeLists.txt
+++ b/decNumber/CMakeLists.txt
@@ -31,6 +31,8 @@ target_include_directories(decNumber
 
 set_target_properties(decNumber
     PROPERTIES
+        VERSION "3.68"
+        SOVERSION "3.68"
         PUBLIC_HEADER "${DEC_PUB_HEADERS}")
 
 # ---


decNumber/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ target_include_directories(decNumber
 
 set_target_properties(decNumber
     PROPERTIES
+        VERSION ${PROJECT_VERSION}
```
nirosys, line 34: We're using version 3.68 of decNumber, I'd prefer if we used that rather than the ion-c project version.
```suggestion
        VERSION "3.68"
```

decNumber/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ target_include_directories(decNumber
...
+        SOVERSION ${PROJECT_VERSION}
```
nirosys, line 35:
```suggestion
        SOVERSION "3.68"
```

nirosys (changes requested):
Awesome, thank you for the PR! I just have one minor request.

nirosys (approved):

PR #329, opened 2023-09-06
Add Benchmark CLI for format comparison and perf tracking

nirosys:
> NOTE: This PR is currently in draft while I put together some notes about the CLI, and do a quick Pre-review review.

*Issue #, if available:*

*Description of changes:*

#### Google Test
This PR changes the googletest usage a little. This was primarily to support google-benchmark's googletest dependency, but resulted in further adjustments to supply users with the ability to:
* Bring their own google test version without colliding with ion-c's version. Just import googletest however is needed so that the `gtest_main` target is available, and the ion-c build will disable its own import.
* Disable ion-c unit tests from building. Turn the `IONC_BUILD_TEST` flag `off` in order to stop the ion-c build from building ion-c unit tests.

#### Build Type: Profiling
This PR's primary purpose is to add a new CLI to allow for benchmarking for performance baselining, analysis, and improvement quantification. In order to produce builds where we have optimized builds, but still have enough debug information to generate flame graphs, and use profilers effectively a new build type was added to do just that. Rather than building with `-DCMAKE_BUILD_TYPE=Release`, a user can now use `-DCMAKE_BUILD_TYPE=Profiling` and cmake will produce a build that has debug information and optimization passes.

### Addition of IonCBench
The primary purpose of this PR is to add the ion-c benchmark CLI. In order to build this tool the user can set the CMake variable `IONC_BENCHMARKING_ENABLED` to `ON`. This is done by default when using the `Profiling` build type. The tool is intended to provide a similar set of features to the other benchmark CLIs found for ion-java, and ion-python.

Currently, the tool allows the user to provide their own data, and perform both full deserialization and full serialization of that data using ion-c (in text mode, or binary),  MsgPack (MsgPack-C), JSON (yyjson, and json-c), and CBOR (libcbor).

This PR is the first iteration of the ion-bench tool, and provides functionality for measuring CPU time, and other CPU metrics such as instruction counts (as long as it is built with libpfm). The tool currently supports a single benchmark run per invocation. Each benchmark run requires an input dataset, a specified supported implementation, and the requested benchmark to run. Currently the tool supports two benchmarks:
* read_all - Where all data is read, and materialized, from the input dataset.
* write_all - Where a flattened representation of the data is read from the input dataset, and re-written using the same format. The timing information is only for the write of the dataset.

In both benchmarks timing does not include the IO to get the input dataset into memory.

The expectation is that more benchmarks will be added, along with more format implementations, in order to compare runtime, memory usage, and data size, between ion and other data formats.

---


nirosys:
Build checks failed due a couple issues. One where gtest isn't linking, which doesn't reproduce on my system. Another is cmake acting like it hasn't known what C++17 is since 3.8.. continuing to dig.

nirosys:
The unit test crash that is occurring with the amazonlinux:1 gcc72 build isn't reliably reproducible locally. The most recent commit should address the memleak that was identified by the ubuntu clang build, but I'm not sure if it is related. Will continue digging if not.

nirosys:
Ok. I've reproduced the crash that GHA is seeing, and it's pretty awesome. /s

All of my attempts to reproduce the issue failed. I combed through the package versions, and made sure the docker digest SHA reported by GitHub matched the image I was using locally. Everything lined up but for some reason the issue would not reproduce. Until I piped the output of act, into `tee` so I could grep through the output to make sure there wasn't any pointers in the haystack of warnings ion-c produces. As soon as the job ran with this new pipe, it crashed with the same error GHA produces.

I was also able to reproduce the issue by simply redirecting the output of the unit tests into a file, which made it possible to easily run it under gdb. I had known from the output of the GHA crash that the issue was triggering in the test_ion_decimal.cpp tests, specifically `WriteAllValues`, but had no reason to see an issue. Debugging without the redirect showed the values within the function to be sane, and no error occurred during the free.

With the redirect, I'm guessing some things have shifted on the stack, and the uninitialized `ion_decimal` defined within the WriteAllValues test ends up lining up with data that makes the `ION_DECIMAL`'s `type` field contain the value for `ION_DECIMAL_TYPE_NUMBER`.  This results in the only code path that tries to free the decimal's `value.num_value` buffer, which is also initialized with random stack data and cannot be free'd.

nirosys:
Putting this into review. There may be an issue with ion-test-drivers still looking into that, but this PR should be good to start moving forward.

tools/ion-bench/src/ion/reading.h
```diff
@@ -0,0 +1,200 @@
...
+
+            ion_reader_has_any_annotations(_reader, &has_anns);
+            if (has_anns == TRUE) {
+               ion_reader_get_annotation_count(_reader, &count);
+               ION_STRING *syms = new ION_STRING[count];
+               ion_reader_get_annotations(_reader, syms, count, &count);
+               for (int i=0; i < count; i++) {
+                  annot.push_back(std::string((char *)syms[i].value, syms[i].length));
+               }
+               delete[] syms;
```
tgregg, line 107: Should we use `ion_reader_get_an_annotation` in the loop and avoid copying into a temporary array?

tools/ion-bench/src/ion/writing.h
```diff
@@ -0,0 +1,237 @@
...
+      WRITE_ION_TYPE(writer, tid_STRING_INT, std::string, val);
+      WRITE_ION_TYPE(writer, tid_INT_INT, int64_t, val);
+      WRITE_ION_TYPE(writer, tid_FLOAT_INT, double, val);
+      WRITE_ION_TYPE(writer, tid_DECIMAL_INT, IonDecimalPtr, val);
+      WRITE_ION_TYPE(writer, tid_BOOL_INT, bool, val);
+      WRITE_ION_TYPE(writer, tid_SYMBOL_INT, Symbol, val);
+      WRITE_ION_TYPE(writer, tid_TIMESTAMP_INT, std::shared_ptr<ION_TIMESTAMP>, val);
+      default:
+         printf("Attempt to write unknown type: %d\n", val.tpe);
+         break;
```
tgregg, line 231: Are we missing support for `null.null`, blob, and clob?

tools/ion-bench/src/json/yy.h
```diff
@@ -0,0 +1,319 @@
...
+               stats.num_nulls++;
+               break;
+            case YYJSON_TYPE_ARR: {
+               stats.num_lists++;
+               auto valstats = process_list(val);
+               stats += valstats;
+               break;
+            }
+            case YYJSON_TYPE_BOOL:
+               stats.num_bools++;
```
tgregg, line 98: Should we retrieve the value of the boolean?

tools/ion-bench/src/main.cpp
```diff
@@ -0,0 +1,260 @@
...
+};
+
+template<size_t N>
+data_name(char const (&)[N]) -> data_name<N - 1>;
+
+/*
+ * Deserialization benchmark.
+ */
+// template <typename T, data_name dataset, Format format>
+// void deser(benchmark::State &state) {
```
tgregg, line 65: Looks like this stuff moved to benchmarks.h. Let's delete the commented code if it's not needed.

tgregg (approved):
Nice! Could you add some example output from the tool to the PR description?

.gitmodules
```diff
@@ -1,7 +1,21 @@
 [submodule "ion-tests"]
 	path = ion-tests
 	url = https://github.com/amazon-ion/ion-tests.git
-[submodule "googletest"]
-	path = test/googletest
+[submodule "tools/ion-bench/deps/libcbor"]
```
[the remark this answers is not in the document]
nirosys: Yea, not sure what we can do here. I might be able to wire it into the cmake config. I'll dig into that, and see what options we have.

tools/ion-bench/src/ion/reading.h
nirosys: Good call. I'll follow up with a PR to change that.

tools/ion-bench/src/ion/writing.h
nirosys: Agh, yea. I'll include those in the follow-up PR.

tools/ion-bench/src/json/yy.h
nirosys: Yea, I need to wire up a dependency on the bool so it doesn't get optimized out. For yyjson specifically getting a bool is a memory access, and a couple bit operations, so there aren't any side effects to keep the call around. I'll address it in the follow-up.

nirosys:
Thank you! I'm going to push up a new commit with the commented code, and typos, fixed, and get this merged. Then I'll PR the regression workflow, and follow up with the code changes discussed above. Unless anyone has an argument against that.

tgregg (approved):

PR #332, opened 2023-10-17
Add performance regression testing workflow

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR adds a GHA workflow that builds both the HEAD of the target branch, as well as the PR's revision, and runs the ion-bench tool to generate benchmark results for a set of 200KiB binary ion datasets.

This PR adds only the read (deserialize_all) benchmark to the workflow. An update to add the write (serialize_all) benchmark will follow.

The workflow uses google-benchmark's `compare.py` which performs a null-hypothesis check agains the data to determine whether the results from both benchmark runs are statistically similar or not. This test is currently not used for the validation. This is primarily because I need to understand the tuning of the alpha parameter, as well as what to expect from the test in a noisy environment like GHA.

Example Run: [here](https://github.com/amazon-ion/ion-c/actions/runs/6545960025/job/17775463662?pr=332)

---

### DIFF
--- a/.github/workflows/performance-regression.yml
+++ b/.github/workflows/performance-regression.yml
@@ -12,12 +12,6 @@ jobs:
       CC: 'clang'
       CXX: 'clang++'
     steps:
-      - name: Install Dependencies
-        run: |
-          apt-get update
-          apt-get install git cmake clang-14 python3-pip -y
-          [[ -e /usr/bin/cmake ]] || ln -s `which cmake3` /usr/bin/cmake
-
       - name: Get Data Generator
         uses: actions/checkout@v2
         with:
--- a/.github/workflows/performance-regression.yml
+++ b/.github/workflows/performance-regression.yml
@@ -1,0 +1,146 @@
+name: Performance Regression Detection
+on:
+  pull_request:
+    branches: [ master ]
+jobs:
+  detect-regression:
+    name: Detect Regression
+    runs-on: ubuntu-latest
+    strategy:
+     fail-fast: true
+    env:
+      CC: 'clang'
+      CXX: 'clang++'
+    steps:
+      - name: Install Dependencies
+        run: |
+          apt-get update
+          apt-get install git cmake clang-14 python3-pip -y
+          [[ -e /usr/bin/cmake ]] || ln -s `which cmake3` /usr/bin/cmake
+
+      - name: Get Data Generator
+        uses: actions/checkout@v2
+        with:
+          repository: amazon-ion/ion-data-generator
+          ref: main
+          path: ion-data-generator
+
+      - name: Build Ion Data Generator
+        run: cd ion-data-generator && mvn clean install
+
+      - name: Generate Data
+        env:
+          jar_file: ion-data-generator/target/ion-data-generator-1.0-SNAPSHOT.jar
+          schema_dir: ion-data-generator/tst/com/amazon/ion/workflow
+        run: |
+          mkdir -p testData
+          # Generate approximately 200KB of data for each dataset, so that we can expect similar orders of magnitude for
+          # our threshold.
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema01.isl testData/realWorldDataSchema01.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema02.isl testData/realWorldDataSchema02.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema03.isl testData/realWorldDataSchema03.10n
+
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/nestedList.isl testData/nestedList.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/nestedStruct.isl testData/nestedStruct.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/sexp.isl testData/sexp.10n
+
+      - name: Fetch PR Candidate
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          path: candidate
+
+      - name: Build PR Candidate
+        run: |
+          mkdir -p candidate/build/profiling && cd candidate/build/profiling
+          cmake -DCMAKE_BUILD_TYPE=Profiling -DIONC_BUILD_TESTS=OFF ../..
+          make clean && make IonCBench
+
+      - name: Fetch PR Baseline
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+          submodules: recursive
+          path: baseline
+
+      - name: Build PR Baseline
+        run: |
+          mkdir -p baseline/build/profiling && cd baseline/build/profiling
+          cmake -DCMAKE_BUILD_TYPE=Profiling -DIONC_BUILD_TESTS=OFF ../..
+          make clean && make IonCBench
+
+      # This step runs benchmarks for the current ion-c repo.
+      - name: 'Benchmark: Baseline'
+        env:
+          cli_path: baseline/build/profiling/tools/ion-bench/src/IonCBench
+        run: |
+          $cli_path -b deserialize_all -l ion-c-binary \
+                    --benchmark_context=uname="`uname -srm`" \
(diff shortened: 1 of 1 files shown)


.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,146 @@
+name: Performance Regression Detection
+on:
+  pull_request:
+    branches: [ master ]
```
linlin-s, line 4: We might want to trigger the workflow only when the source code of `ion-c` has been changed. 
```
on:
pull_reuqest: 
  paths:
        - 'ionc/*'
```

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,146 @@
...
+          mkdir -p testData
+          # Generate approximately 200KB of data for each dataset, so that we can expect similar orders of magnitude for
+          # our threshold.
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema01.isl testData/realWorldDataSchema01.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema02.isl testData/realWorldDataSchema02.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/realWorldDataSchema03.isl testData/realWorldDataSchema03.10n
+
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/nestedList.isl testData/nestedList.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/nestedStruct.isl testData/nestedStruct.10n
+          java -jar $jar_file generate -S 200000 --input-ion-schema $schema_dir/sexp.isl testData/sexp.10n
```
linlin-s, line 39: Nitpick: We might prefer to use a for loop to avoid repeated logic.
```
for test in nestedStruct nestedList sexp realWorldDataSchema01 realWorldDataSchema02 realWorldDataSchema03
do
java -jar $jar_file generate -S ${{env.data_size}} --input-ion-schema $schema_dir/${test}.isl testData/${test}.10n
done

```

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,146 @@
...
+      CXX: 'clang++'
+    steps:
+      - name: Install Dependencies
+        run: |
+          apt-get update
+          apt-get install git cmake clang-14 python3-pip -y
+          [[ -e /usr/bin/cmake ]] || ln -s `which cmake3` /usr/bin/cmake
+
+      - name: Get Data Generator
+        uses: actions/checkout@v2
```
linlin-s, line 16: Could we pin to a specific commit instead of commit hash tag? Tags can be updated and re-pointed, so the code referenced by a tag can change. However the commit SHA represents the exact snapshot of the code at a point in time. 

The latest version of actions/checkout is recommended. `actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`
Same suggestion to other action/checkout call.

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,146 @@
...
+      # the 'compare.py' script provided by google-benchmark, which will compare the results of the benchmarks to
+      # the results of the baseline benchmarks from the previous step.
+      #
+      # The compare script uses the defined 'alpha' environment variable to perform a null-hypothesis test,
+      # which is used to determine whether the two sets of benchmark times come from the same distribution.
+      - name: 'Benchmark: PR Candidate'
+        env:
+          compare: candidate/tools/ion-bench/deps/google-benchmark/tools/compare.py
+          cli_path: candidate/build/profiling/tools/ion-bench/src/IonCBench
+          alpha: 0.03
```
linlin-s, line 98: I'd like to know if alpha value can be adjusted based on the accuracy of regression-detection workflow?

linlin-s:
Would you mind adding the sample run of this workflow to the description?

.github/workflows/performance-regression.yml
nirosys: Good call, adding.

.github/workflows/performance-regression.yml
nirosys: Yea.. I had originally broken it out because what I had taken from either ion-python's, or ion-java's
workflow had `-S 100` which resulted in some wildly different dataset sizes. I think the sexp.10n was 195bytes, which definitely wasn't large enough to overcome the normal runtime noise (5% diff ended up being ~195ns on my system). Once I saw what `-S` does, and realized what was going on, I didn't re-loop it. I'll tidy that up.

nirosys:
> Would you mind adding the sample run of this workflow to the description?

@linlin-s, I'm not sure what you mean. The workflow ran as part of the checks, so all of the output is available there. The full output is a bit much for the PR description, do you mean just the output of the `compare.py`?

.github/workflows/performance-regression.yml
linlin-s: Thanks for the explanation.

linlin-s:
> > Would you mind adding the sample run of this workflow to the description?
> 
> @linlin-s, I'm not sure what you mean. The workflow ran as part of the checks, so all of the output is available there. The full output is a bit much for the PR description, do you mean just the output of the `compare.py`?

Apologize for the confusion. What I mean by sample run is already included in the checks. We can track the current workflow run in the checks since it is triggered by `pull_request` to `master`. After we update the trigger event, we might have to manually trigger the workflow to see the results since the workflow will be triggered only when the source code changed. In this case it would be easier to review if we could include the hyperlink of the updated workflow run.

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,146 @@
...
+      # Currently, this check looks at the generated comparison of the MEAN of each benchmarks' CPU time. We
+      # do this for now, rather than use the null-hypothesis results, until we get a better understanding of
+      # how the timings will be in GHA.
+      - name: 'Check Results'
+        env:
+          # Threshold Percentage, currently 5%.
+          threshold_perc: 5
+        run: |
+          echo "Printing Results"
+          RESULTS=$(cat results.json | jq '.[] | select(.run_type == "aggregate" and (.name | endswith("_mean"))) | {name:.name,cpu_time_perc_diff:(.measurements[0].cpu*100)}|select(.cpu_time_perc_diff > '"${threshold_perc}"')')
```
linlin-s, line 134: Approve this PR with one question here: When the `cpu_time_perc_diff` > 0, it represents the candidates CPU run time is bigger than the baseline CPU run time right? If the value is bigger than the threshold value we will fail the workflow with regression detected.

linlin-s (approved):

.github/workflows/performance-regression.yml
nirosys: Yes, exactly. The way the `compare.py` generates its output, we end up with data like this:
```json
  {
    "name": "realWorldDataSchema03.10n_mean",
    "label": "",
    "measurements": [
      {
        "real_time": 12452149.371052643,
        "cpu_time": 12451057.894736838,
        "real_time_other": 12473477.750893278,
        "cpu_time_other": 12470339.46428573,
        "time": 0.0017128271758622316,
        "cpu": 0.0015485888598303227
      }
    ],
    "time_unit": "ns",
    "run_type": "aggregate",
    "aggregate_name": "mean",
    "utest": {}
  },
```
This represents the baseline (cpu\_time) and candidate's (cpu\_time\_other) mean (of 20 runs) wallclock, and cpu time for the realWorldDataSetSchema03 test. The `cpu` field represents the relative difference (`(candidate - baseline)/(baseline)`) between the candidate and the baseline. Likewise, the `time` field represents the relative wallclock difference. Since the relative difference is a percentage of the baseline, I multiply it by 100 it so we can configure the threshold with whole values. Any positive value is a results of the candidate's time being larger than the baseline's.

The check step, passes the JSON output through that jq query. The JSON output from the compare.py, contains both the raw results from the candidate run, but also the comparisons to the baseline like I mentioned above. The jq query looks for any aggregate comparisons, for the mean, and gathers only results that exceed the threshold. If there are no values, the `RESULTS` variable will be an empty string, otherwise it will contain JSON that includes the name and CPU %diff for each failing run. The step then just checks to see if `RESULTS` is empty or not, when deciding if the step should fail or not. If it's not, it will also use jq to print out the name and %diff for each of the tests that exceed the threshold.

Hah.. typing this up made me realize I missed the `aggregate_name` field. I could make the jq query a little friendlier. I'll follow up with that in a later PR.

tgregg (approved):

PR #343, opened 2024-07-29
Fix MacOS build & test

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
This PR removes a previously added configuration that forced the use of Xcode 14.1 that was put in place due to an issue with gcc that occurred in 14.0, while Github runners were defaulting to 14.0. Xcode 14.0 has been removed from the runner images, so the build is now failing.

After this PR, builds will return to using the default Xcode.

> NOTE: Both amazon linux builds are failing due to a node version update (PR incoming); ion-test-driver is failing and a fix has been posted in #344.
---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -103,12 +103,6 @@ jobs:
           fetch-depth: 50
       - name: Create git config # Fixes an issue where git refuses to work due to dubious permissions.
         run: git config --global --add safe.directory $GITHUB_WORKSPACE
-      # This step should fix an issue with building on macos with gcc, and xcode 14.0,
-      # and should be removed once 14.1 is made default.
-      - name: Fix for XCode 14.0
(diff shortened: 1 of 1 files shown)


nirosys:
The failed amazonlinux 1 & 2 builds are a known issue (as of today). GHA is now defaulting to Node20 for running actions like checkout, and node20 is built with a newer version of glibc than is available in either amazonlinux images. There is a potential fix, that will work until node16 is removed from the image that I'll PR shortly. I'll probably remove amazonlinux:1 in that PR, since it is EOL'd as of Jan 1 2024.

ion-test-driver also failed, which is the same issue we saw previously in ion-java, will follow up with a PR for that as well.

zslayton (approved):

jobarr-amzn (approved):

PR #353, opened 2024-11-08
Update MacOS build to use GCC 14

nirosys:
*Issue #, if available:* actions/runner-images#10213

*Description of changes:*
Recently GCC 11 was removed from all MacOS images. This PR updates our build & test to use GCC 14.

Might be worth seeing if we can auto-populate the alias so we don't have to keep this in sync.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
             compiler: { cc: 'gcc', cxx: 'g++' }
           - image: 'macos-latest'
             toolchain: gcc
-            compiler: { cc: 'gcc-11', cxx: 'g++-11' }
+            compiler: { cc: 'gcc-14', cxx: 'g++-14' }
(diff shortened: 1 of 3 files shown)


nirosys:
More reason to double down on getting rid of all of the warnings that ion-c produces, and enforcing a zero warning policy. Clang 15 defaults to treating this as a warning. GCC 14 raises it as an error.

nirosys:
Also:
```
/Users/glitch/Code/ion-c/ionc/ion_internal.h:112:71: error: passing argument 2 of 'ion_stream_read_byte' from incompatible pointer type [-Wincompatible-pointer-types]
  112 |                                   IONCHECK(ion_stream_read_byte((xh), &(xb)));  \
      |                                                                       ^~~~~
      |                                                                       |
      |                                                                       uint64_t * {aka long long unsigned int *}
```

linlin-s (approved):

Issue #346, opened 2024-07-30
Update Amazon Linux in GHA Workflows


nirosys:
We currently have Amazon Linux 1 & 2 in our GHA workflows for building and testing ion-c. AL 1 has been EOL since the beginning of the year, and AL 2 has ~1 year left.

AWS has an explanation of the intended lifecycle for Amazon Linux [here](https://aws.amazon.com/blogs/aws/amazon-linux-2023-a-cloud-optimized-linux-distribution-with-long-term-support/). We should expect AL2023 to be actively supported until 03-15-2025, at which point it will enter maintenance mode. AL2025 should be expected sometime in early 2025, with a tech preview prior to GA.

GH is has changed the GHA runners to default to node20, which is built against a more recent version of glibc than AL1 & AL2 supports. We're able to force the use of node16 for now, until GH completely removes node16 from the image. If we want to keep AL2 in the workflows until it is EOL'd we'll probably need to find a solution for when node16 is completely removed.

nirosys:
[Looks like node16 was removed](https://github.com/amazon-ion/ion-c/actions/runs/12696655845/job/35391168747) from GHA images. 🎉

nirosys:
Node 18 is apparently not [supported by the runner](https://github.com/actions/runner/blob/main/src/Runner.Common/Util/NodeUtil.cs#L9).. which means the Node 18 that is present on the image cannot be used by actions.. I guess this is the forcing function to remove amazonlinux 1 & 2 from our tests. We can add amazonlinux 2023 in their place.

PR #357, opened 2025-01-10
Update Amazon Linux version(s) used in Build & Test workflow

nirosys:
*Issue #, if available:* #346

*Description of changes:*
Recently Github bumped up the minimum version of Node that could be used for github actions to Node 20. This change had the side effect of also bumping up the minimum supported version of glibc for actions to 2.27.

Prior to this PR, we used both Amazon Linux 1 and 2 for our build and test workflow. Unfortunately, AM1 uses glibc 2.17, and AM2 uses glibc 2.26.

Amazon Linux 1 was EOL'd Dec 31, 2023, and stopped receiving security updates Jan 1, 2024. After multiple extensions, Amazon Linux will be EOL'd June 30, 2026. Amazon's recommendation is to use Amazon Linux 2023, which will have full support until 2028.

This PR removes both Amazon Linux 1 and 2, and adds 2023 in their place. Once AL 2025 is available, we should add that as well.

In addition, this PR also cleans up the GHA job by fixing some `actionlint` warnings, which includes updating to checkout v4.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,22 +16,12 @@ jobs:
         image: ['amazonlinux:2023']
         toolchain: [ "gcc", "clang" ]
         include:
-          - image: amazonlinux:2
+          - image: amazonlinux:2023
             toolchain: gcc
             compiler: { cc: 'gcc', cxx: 'g++', packages: 'gcc gcc-c++' }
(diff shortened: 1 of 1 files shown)


jobarr-amzn (approved):
Thanks for also catching/cleaning the `cmake3` linking logic :)

Issue #358, opened 2025-01-24
Error building unit tests with MSVC


nirosys:
While adding a windows target our ion-c CI/CD actions, it was discovered that unit tests are failing to build under MSVC due to the use of designated initializers (a C++20 feature) that GCC & Clang must allow without requiring C++20.

```
D:\a\ion-c\ion-c\test\test_ion_stream.cpp(294,9): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\ion-c\ion-c\build64\test\all_tests.vcxproj]
```

I'm going to look into disabling any GCC extensions that this may have supported this. We're requiring C++14 for unit tests, and that should be strict.

nirosys:
Also ran into:

```
error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MTd_StaticDebug' doesn't match value 'MDd_DynamicDebug' in gather_vectors.obj [D:\a\ion-c\ion-c\build64\test\all_tests.vcxproj]
```

Which was a result of refactoring how dependencies were handled. Google test was included at the top level of the CMake hierarchy, this lead to it being added to the build tree long before the `gtest_force_shared_crt` option was set in the `test/CMakeList.txt`. This caused a linkage issue since google test defaults to static linking, and MSVC does not.

nirosys:
Fixed with #359.

Issue #255, opened 2021-10-08
Replace AppVeyor CI run with Github Actions


popematt:
Very nice. While we're here: does Github Actions support Windows now? We still have an AppVeyor run for Windows, but it we can move that to Github Actions too, that would be great.

_Originally posted by @tgregg in https://github.com/amzn/ion-c/pull/254#pullrequestreview-775340772_

nirosys:
Completed with #359.

PR #359, opened 2025-01-25
Add windows build to Build & Test workflow.

nirosys:
*Issue #, if available:* #358, #255

*Description of changes:*
This PR adds a windows target to our Build & Test workflow, and addresses a couple issues that got in the way of successful windows builds while testing for this PR.

Some more information about the issues that were fixed can be found in #358. The first issue was due to GCC/Clang allowing named initializers (which are now a C++20 feature, but supported in C99) in C++, while MSVC does not.

The second issue was a result of refactoring dependencies a while ago, which resulted in google test being imported into the build tree earlier than a key option (`gtest_force_shared_crt`) to not be set until after google test was imported, resulting in it not taking effect, and causing a linking error. I've moved that setting up to the top-level CMakeLists.txt.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,7 +117,6 @@ jobs:
         with:
           submodules: recursive
           fetch-tags: true
-          fetch-depth: 50
       - name: Build Debug
         run: |
           mkdir build64
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -123,11 +123,13 @@ jobs:
           mkdir build64
           cmake -G "Visual Studio 17 2022" -A x64 -S . -B "build64"
           cmake --build build64 --config Debug
-      - name: Test debug
+      - name: Test Debug
         run: 'build64\test\Debug\all_tests.exe'
       - name: Build Release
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -108,6 +108,29 @@ jobs:
...
+  test_windows:
+    name: "Test (windows-latest, msvc)"
+    runs-on: windows-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-tags: true
+          fetch-depth: 50
```
jobarr-amzn, line 120:
```suggestion
      - name: Checkout Code
        uses: actions/checkout@v4
        with:
          submodules: recursive
          fetch-tags: true
```

jobarr-amzn (approved):

.github/workflows/main.yml
nirosys: Ah yea, this can be removed now. Ty for catching that.

jobarr-amzn (approved):

PR #368, opened 2026-05-09
Hardens arithmetic against overflow.

tgregg:
## Summary

- Fix integer overflow in pool allocator size computation that could produce undersized allocations when `min_needed` is near `INT32_MAX`
- Fix integer overflow in `ion_int_from_bytes` / `ion_int_from_abs_bytes` and binary reader int helpers where `byte_count * 8` wraps `int32_t` for large inputs
- Fix `ion_int_is_null` incorrectly gating its output write on an uninitialized value
- Replace pointer-arithmetic bounds checks in `_ion_alloc_with_owner_helper` with length-based comparisons that cannot wrap
- Add early validation in `_ion_reader_binary_local_read_length` rejecting lengths exceeding `INT32_MAX`
- Validate `max_id` accumulation in symbol table imports to prevent overflow past `INT32_MAX`
- Guard multiplication overflow in `_ion_index_grow_array` and `_ion_index_make_room`
- Guard symbol table index doubling against overflow
- Reject binary Ion field SIDs exceeding `INT32_MAX`
- Widen bit-count arithmetic in `_ion_int_highest_bit_set_helper` and `_ion_int_is_high_bytes_high_bit_set_helper` to prevent overflow
- Add early rejection of oversized decimal digit strings in `ion_int_from_chars`
- Guard `_ion_int_extend_digits` against allocation size overflow
- Fix `ion_string_strdup` overflow when `length == INT32_MAX`
- Add `ReleaseSan` build type (NDEBUG + ASan/UBSan) and corresponding CI job to catch issues masked by debug-only assertions

## Testing

All existing tests pass in both Release and Debug configurations (2,994 tests).

New unit tests in `test/test_ion_overflow.cpp` verify:
- The pool allocator rejects requests near `INT32_MAX`, negative sizes, and zero sizes rather than returning undersized blocks
- `ion_int_from_abs_bytes` and `ion_int_from_bytes` return `IERR_INVALID_ARG` when the byte count is large enough that bit-count arithmetic would overflow `int32_t`
- The binary reader returns errors for values whose declared length exceeds the actual stream data
- Symbol table imports with `max_id` values that sum past `INT32_MAX` are rejected
- Index array growth rejects `new_count * entry_size` products that overflow `int32_t`
- Index bucket threshold computation rejects values where `* 128` overflows
- Binary reader rejects struct field SIDs exceeding `INT32_MAX`
- `_ion_int_extend_digits` rejects `digits_needed` values where `* sizeof(II_DIGIT)` overflows
- `ion_int_highest_bit_set` returns correct results without overflowing internal arithmetic
- `ion_int_to_bytes` handles large byte counts without overflowing the high-bit helper
- `ion_int_from_chars` rejects decimal strings longer than `MAX_SIZE / 4` digits
- `ion_string_strdup` returns NULL for strings with `length >= MAX_SIZE`

A new `ReleaseSan` build configuration compiles with `NDEBUG` defined alongside AddressSanitizer and UndefinedBehaviorSanitizer. This catches correctness issues that only manifest in production-like builds where debug assertions are stripped. A corresponding CI job runs the full test suite under this configuration on every push and pull request.

### DIFF
--- a/build-release-sanitizers.sh
+++ b/build-release-sanitizers.sh
@@ -1,0 +1,17 @@
+#!/bin/sh --
+
+set -x
+
+export LD="$CXX"
+
+if ! [ -x "$(command -v cmake)" ]; then
+  echo 'Error: cmake is not installed.' >&2
+  exit 1
+fi
+
+mkdir -p build/release-san && cd build/release-san
+cmake \
+  -DCMAKE_BUILD_TYPE=ReleaseSan \
+  ${CMAKE_FLAGS} \
+  ../..
+make clean && make -j"$(nproc || sysctl -n hw.ncpu)"
--- a/ionc/ion_allocation.c
+++ b/ionc/ion_allocation.c
@@ -119,15 +119,15 @@ void *_ion_alloc_with_owner_helper(ION_ALLOCATION_CHAIN *powner, SIZE request_le
         // check for space in blocks that are "in progress", starting with the
         // owner block, we only need to do this if we aren't already directed
         // to allocate a new block by force_new_block
-        next_ptr = pblock->position + length;
-        if (next_ptr > pblock->limit) {
+        SIZE remaining = (SIZE)(pblock->limit - pblock->position);
+        if (length < 0 || length > remaining) {
             pblock = powner->head;
             if (pblock == NULL) {
                 force_new_block = TRUE;
             }
             else {
-                next_ptr = pblock->position + length;
-                if (next_ptr > pblock->limit) {
+                remaining = (SIZE)(pblock->limit - pblock->position);
+                if (length > remaining) {
                     force_new_block = TRUE;
                 }
             }
@@ -159,9 +159,10 @@ void *_ion_alloc_with_owner_helper(ION_ALLOCATION_CHAIN *powner, SIZE request_le
             pblock->next = powner->head;
             powner->head = pblock;
         }
-        next_ptr = pblock->position + length;
-        assert(next_ptr <= pblock->limit); // we better have room at this point
+        assert(length <= (SIZE)(pblock->limit - pblock->position));
     }
+
+    next_ptr = pblock->position + length;
 
     // if we have a block here, there's enough room in it
     // save the current position ptr for our caller, and move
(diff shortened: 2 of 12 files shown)


ionc/ion_allocation.c
```diff
@@ -119,15 +119,13 @@ void *_ion_alloc_with_owner_helper(ION_ALLOCATION_CHAIN *powner, SIZE request_le
         // check for space in blocks that are "in progress", starting with the
         // owner block, we only need to do this if we aren't already directed
         // to allocate a new block by force_new_block
-        next_ptr = pblock->position + length;
-        if (next_ptr > pblock->limit) {
+        if (length < 0 || length > (SIZE)(pblock->limit - pblock->position)) {
```
jobarr-amzn: We are using `pblock->limit - pblock->position` repeatedly, would `remaining` be a reasonable name for this?

ionc/ion_symbol_table.c
```diff
@@ -2200,6 +2203,9 @@ iERR _ion_symbol_table_index_insert_helper(ION_SYMBOL_TABLE *symtab, ION_SYMBOL
     if (adjusted_sid > symtab->by_id_max) {
         // the +1 is because sid's are 1 based (so we're losing the 0th slot, and need 1 extra entry)
         old_count = (symtab->by_id_max + 1);
+        if (old_count > MAX_SIZE / DEFAULT_SYMBOL_TABLE_SID_MULTIPLIER) {
```
jobarr-amzn, line 2207:
```suggestion
    // can't import so many symbols that we overflow int for symbol table size
    if (import_max_id > MAX_SIZE - symtab->max_id) {
```

build-release-sanitizers.sh
```diff
@@ -0,0 +1,17 @@
...
+  echo 'Error: cmake is not installed.' >&2
+  exit 1
+fi
+
+mkdir -p build/release-san && cd build/release-san
+cmake \
+  -DCMAKE_BUILD_TYPE=ReleaseSan \
+  ${CMAKE_FLAGS} \
+  ../..
+make clean && make -j"$(nproc || sysctl -n hw.ncpu)"
```
jobarr-amzn, line 17: I'm a little surprised that this isn't an option to `make`. Like "`make`, and use as much CPU as you can please".

jobarr-amzn (dismissed):

ionc/ion_symbol_table.c
tgregg: Added a comment

jobarr-amzn (approved):

PR #369, opened 2026-05-11
Additional overflow hardening.

tgregg:
## Summary

- Fix heap overflow in `_ion_scanner_read_as_base64` where post-decrement underflow causes the decode loop to write past the caller's buffer
- Add recursion depth limit to `_ion_scanner_skip_container` to prevent stack exhaustion from deeply nested text Ion containers
- Fix out-of-bounds read and broken logic in `_ion_int_from_chars_helper` null literal parsing that used `strncmp` on non-NUL-terminated strings

## Testing

New unit tests in `test/test_ion_overflow.cpp` verify:
- `ion_reader_read_lob_partial_bytes` with `buf_max % 3 == 2` does not write past the provided buffer when decoding base64 blobs
- Text Ion input with 1000-deep nested containers returns an error rather than recursing unboundedly when the scanner skips a container value
- `ion_int_from_chars` correctly accepts "null" (length 4) and "null.int" (length 8), rejects partial matches like "n" (length 1), and never reads past the declared string length

### DIFF
--- a/test/test_ion_overflow.cpp
+++ b/test/test_ion_overflow.cpp
@@ -435,7 +435,7 @@ TEST(IonBinaryReader, RejectsOversizedSymbolTableString) {
 
 TEST(IonInteger, FromCharsHandlesNullLiterals) {
     // _ion_int_from_chars_helper uses strncmp on non-NUL-terminated strings
-    // and has broken || logic. After the fix, "null" and "null.int" should
+    // and had broken || logic. After the fix, "null" and "null.int" should
     // be accepted, "n" alone should be rejected, and no OOB read should occur.
 
     ION_INT *iint = NULL;
--- a/test/test_ion_overflow.cpp
+++ b/test/test_ion_overflow.cpp
@@ -507,9 +507,9 @@ TEST(IonTextReader, DeeplyNestedContainersDoNotCrash) {
 }
 
 TEST(IonTextReader, Base64DecodeDoesNotOverflowBuffer) {
-    // _ion_scanner_read_as_base64 has a post-decrement underflow when
-    // buf_max % 3 == 2. The 'remaining' counter wraps to -1 and the
-    // decode loop writes past the caller's buffer.
+    // _ion_scanner_read_as_base64 had a post-decrement underflow when
+    // buf_max % 3 == 2. The 'remaining' counter wrapped to -1 and the
+    // decode loop wrote past the caller's buffer.
     //
     // Text Ion blob: {{ AQID }} which is base64 for bytes 0x01 0x02 0x03
     // We read with buf_max=2 so the 3-byte decode hits the boundary.
--- a/test/test_ion_overflow.cpp
+++ b/test/test_ion_overflow.cpp
@@ -474,10 +474,10 @@ TEST(IonTextReader, DeeplyNestedContainersDoNotCrash) {
     // the inner deeply-nested list. This triggers _ion_scanner_skip_container.
     const int DEPTH = 1000;
     std::string ion_text;
-    ion_text += '[';
+    ion_text += '[  ';
     ion_text += std::string(DEPTH, '[');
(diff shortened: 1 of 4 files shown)


test/test_ion_overflow.cpp
```diff
@@ -432,3 +432,116 @@ TEST(IonBinaryReader, RejectsOversizedSymbolTableString) {
...
+    //
+    // Format: [  [[[...1000 brackets...]]]  ]
+    // We step into the outer list, then call next() which must skip
+    // the inner deeply-nested list. This triggers _ion_scanner_skip_container.
+    const int DEPTH = 1000;
+    std::string ion_text;
+    ion_text += '[';
+    ion_text += std::string(DEPTH, '[');
+    ion_text += std::string(DEPTH, ']');
+    ion_text += ']';
```
jobarr-amzn: Just for fun, changing the content to match the comment because the comment is neatly legible :P

```suggestion
    // Format: [  [[[...1000 brackets...]]]  ]
    // We step into the outer list, then call next() which must skip
    // the inner deeply-nested list. This triggers _ion_scanner_skip_container.
    const int DEPTH = 1000;
    std::string ion_text;
    ion_text += '[  ';
    ion_text += std::string(DEPTH, '[');
    ion_text += std::string(DEPTH, ']');
    ion_text += '  ]';
```

test/test_ion_overflow.cpp
```diff
@@ -432,3 +432,116 @@ TEST(IonBinaryReader, RejectsOversizedSymbolTableString) {
...
+    ASSERT_NE(IERR_OK, err);
+
+    ion_reader_close(reader);
+}
+
+TEST(IonTextReader, Base64DecodeDoesNotOverflowBuffer) {
+    // _ion_scanner_read_as_base64 has a post-decrement underflow when
+    // buf_max % 3 == 2. The 'remaining' counter wraps to -1 and the
+    // decode loop writes past the caller's buffer.
+    //
```
jobarr-amzn: Past tense, because the test is asserting that this behavior no longer exists.

```suggestion
    // _ion_scanner_read_as_base64 had a post-decrement underflow when
    // buf_max % 3 == 2. The 'remaining' counter wrapped to -1 and the
    // decode loop wrote past the caller's buffer.
    //
```

test/test_ion_overflow.cpp
```diff
@@ -432,3 +432,116 @@ TEST(IonBinaryReader, RejectsOversizedSymbolTableString) {
...
+
+    // "null.in" (length 7) should fail — not a complete literal
+    err = ion_int_from_chars(iint, "null.in", 7);
+    ASSERT_EQ(IERR_INVALID_SYNTAX, err);
+
+    ion_int_free(iint);
+}
+
+TEST(IonTextReader, DeeplyNestedContainersDoNotCrash) {
+    // _ion_scanner_skip_container recurses with no depth limit.
```
jobarr-amzn:
```suggestion
    // _ion_scanner_skip_container should not recurse without a depth limit.
```

test/test_ion_overflow.cpp
```diff
@@ -432,3 +432,116 @@ TEST(IonBinaryReader, RejectsOversizedSymbolTableString) {
 
     ion_reader_close(reader);
 }
+
+TEST(IonInteger, FromCharsHandlesNullLiterals) {
+    // _ion_int_from_chars_helper uses strncmp on non-NUL-terminated strings
+    // and has broken || logic. After the fix, "null" and "null.int" should
```
jobarr-amzn:
```suggestion
    // _ion_int_from_chars_helper uses strncmp on non-NUL-terminated strings
    // and had broken || logic. After the fix, "null" and "null.int" should
```

jobarr-amzn (dismissed):

jobarr-amzn (approved):

